PR #4, opened 2021-10-28
Add ParallelTempering

mswwilson:
This needs to be tuned and set up for simulations, but the basic implementation is there.

### DIFF
--- a/examples/PT_afm_heisenberg.jl
+++ b/examples/PT_afm_heisenberg.jl
@@ -1,4 +1,5 @@
 using Sunny
+using MPI
 
 # run using the command: "mpiexec -n [n_procs] julia --project [PT_afm_heisenberg.jl]"
 
--- a/examples/PT_afm_heisenberg.jl
+++ b/examples/PT_afm_heisenberg.jl
@@ -1,0 +1,37 @@
+using Sunny
+
+# run using the command: "mpiexec -n [n_procs] julia --project [PT_afm_heisenberg.jl]"
+
+#######################################################
+#   system setup
+
+# make lattice
+crystal = Sunny.diamond_crystal(a=1.0)
+
+# interactions -- units of K           
+J = 28.28
+interactions = [
+    heisenberg(J, Bond{3}(3, 6, [0,0,0])),
+]
+
+# spin system
+extent = (4,4,4)
+sys = SpinSystem(crystal, interactions, extent)
+sys_size = length(sys)
+rand!(sys)
+
+#######################################################
+#   PT setup
+
+T_min = 22.5
+T_max = 27.5
+
+# temperature schedule: should take index (i) and number of processes (N) and return temperature (T)
+T_sched(i, N) = (10 .^(range(log10(T_min), log10(T_max), length=N)) )[i]
+
+# make replica for PT
+replica = Replica(MetropolisSampler(sys, 1.0, 1))
+
+# run PT
+run!(replica, T_sched; therm_mcs=10000, measure_interval=1, rex_interval=100, max_mcs=10000, bin_size=J)
+
--- a/src/ParallelTempering.jl
+++ b/src/ParallelTempering.jl
@@ -1,4 +1,3 @@
-using MPI
 
 """ 
 Container for sampler type as there is no deriving from concrete types in Julia
--- a/src/ParallelTempering.jl
+++ b/src/ParallelTempering.jl
@@ -27,11 +27,9 @@ function Replica(s::AbstractSampler)
 
     Random.seed!(round(Int64, time()*1000))
 
-    # even rank exch. down when rex_dir==0, up when rex_dir==1
-    nn_ranks = [
-        rank + 2*(rank%2) - 1,
-        rank + 1 - 2*(rank%2)
-    ]
+    # even rank exch. down when rex_dir==1, up when rex_dir==2
+    nn_ranks = [rank-1, rank+1]
+
     # first and last replicas only exch. in one dir.
     for i in 1:2
         if (nn_ranks[i] < 0) || (nn_ranks[i] >= N_ranks)
@@ -39,7 +37,10 @@ function Replica(s::AbstractSampler)
         end
     end
 
-    return Replica(rank, N_ranks, 1, nn_ranks, [-1,-1], s)
+    # start even ranks exchanging down
+    rex_dir = (rank % 2 == 0) ? 1 : 2
+
+    return Replica(rank, N_ranks, rex_dir, nn_ranks, [-1,-1], s)
 end
 
 
@@ -50,14 +51,12 @@ Propose and accept/reject a replica exchange.
 -> Use deterministic even/odd scheme (irreversible).
 """
 function replica_exchange!(replica::Replica)
-    rex_ID = replica.rex_dir+1
-
     # first and last replicas only exch. in one dir.
-    if replica.nn_ranks[rex_ID] < 0
+    if replica.nn_ranks[replica.rex_dir] < 0
         return false
     end
 
-    rex_rank = replica.nn_ranks[rex_ID]
+    rex_rank = replica.nn_ranks[replica.rex_dir]
     E_curr = [running_energy(replica.sampler)]
     rex_accept = [false]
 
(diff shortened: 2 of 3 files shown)


kbarros:
This looks unobtrusive to me, but the only concern I have is about adding more `using` statements to the main Sunny.jl package if they are only applicable to certain use cases. In this case, I'm not sure all users want or need `MPI.jl`. We had a similar problem with `GLMakie`. Matt, what about a loader for the ParallelTempering module that does `@eval using MPI` at run time? Let's discuss.

ColeMiles:
Could you add an example which uses the parallelization / multiple processes? I can't quite figure out how this is intended to be run in the parallel case.

We should also think about how to unify measurements across the single CPU "regular" Sunny code and structs like `WangLandau` and the Parallel Tempering which constructs ensembles of systems. It seems like we shouldn't have every struct re-implement internal code which does measurements, and we also need some way for users to be able to make custom measurements without changing the internal code of these types' methods.

This seems hard enough that we should probably have a meeting about it.

kbarros:
There are several design decision to be made -- I agree we should have a meeting.

mswwilson:
I'm fine with changing where / how MPI is loaded. The file "PT_afm_heiseberg.jl" should be a working example, although the code is more of a skeleton that doesn't make any measurements yet, but simply collects a histogram on each process, performs replica exchanges, and prints the number of accepted exchanges. I run it from the command line like this: "mpiexec -n <nprocs> julia --project PT_afm_heisenberg.jl".

ColeMiles:
Oh I see -- I should have read the comment you provided at the top of the script :laughing:. Thanks.

kbarros:
What are the next steps? I'd like to get these pull requests merged quickly, and Matt can refine later. I guess the main thing is to check whether the `using MPI` statement will be a problem for ordinary users, and if we can instead `@eval using MPI` at run time ?

kbarros:
I think this should be good to merge now.

examples/PT_afm_heisenberg.jl
```diff
@@ -0,0 +1,37 @@
+using Sunny
+
```
ColeMiles, line 2: Now that we have `@require MPI`, I think this script needs `using MPI` at the top to bring `Replica` and `run!` into scope.

src/ParallelTempering.jl
```diff
@@ -0,0 +1,168 @@
...
+
+    # write replica exchange rates to file for each process
+    f = open(@sprintf("P%03d_rex.dat", replica.rank), "w")
+    println(f, "REX accepts (down, up) = (", rex_accepts[1], " ", rex_accepts[2],"), total = ", sum(rex_accepts), " / ", N_rex)
+    close(f)
+
+    # write histogram to file for each process
+    f = open(@sprintf("P%03d_hist.dat", replica.rank), "w")
+    println(f, A)
+    close(f)
```
ColeMiles, line 164: Would it be preferable for all the processes to communicate their measurements back to a master rank, which then writes them all out into a single file?

src/ParallelTempering.jl
```diff
@@ -0,0 +1,168 @@
...
+    # start PT with finite length
+    for i in 1:N_rex
+
+        # measurements between exchanges
+        for j in 1:N_measure
+            # decorrelation steps
+            sample!(replica.sampler)
+
+            # add measurements here 
+            #...
```
ColeMiles, line 141: This could be solved in a future pull. Ideally, a typical user wouldn't have to edit any Sunny. One idea to get arbitrary measurements in from the user:

`Replica` could take a `Function` called `measure`, where calling `measure(system)` returns a `Dict` of measurements mapping a string (the measurement name) to whatever form the measurement takes. `Replica` would then call this function right where you currently have `# add measurements here`, and then somehow keep track of all of the measurements internally in a big `Dict`. You could use `Serialization` to `serialize` it out into a big file at the end.

Users can then get whatever measurements they want by passing a custom measurement function.

Not sure what the pitfalls of this approach would be though.

src/ParallelTempering.jl
ColeMiles: Also: if the analysis of the output data is going to also be done in Julia, it seems like it might be better to use `Serialization` to `serialize` out the `BinnedArray` to make it easy to re-load (and probably uses less storage).

Though, maybe users want to do the analysis with something else, in which case plain-text makes sense. Serialized files also don't transfer between different versions of Julia / probably different versions of packages. Maybe this is a flag to set? Not sure.

src/ParallelTempering.jl
```diff
@@ -0,0 +1,168 @@
...
+    rank = MPI.Comm_rank(MPI_COMM_WORLD)
+    N_ranks = MPI.Comm_size(MPI_COMM_WORLD)
+
+    Random.seed!(round(Int64, time()*1000))
+
+    # even rank exch. down when rex_dir==0, up when rex_dir==1
+    nn_ranks = [
+        rank + 2*(rank%2) - 1,
+        rank + 1 - 2*(rank%2)
+    ]
```
ColeMiles, line 35: This may just be me, but I think it's a bit more interpretable if `nn_ranks` is your left/right neighbor for all ranks, and then you start off even ranks with `rex_dir = 1` and odd ranks with `rex_dir=2`.

Then, in `run!` you flip with `rex_dir = 3 - rex_dir` and you no longer need `rex_ID` in `replica_exchange!`. Might as well stay in the 1-indexing.

examples/PT_afm_heisenberg.jl
```diff
@@ -0,0 +1,37 @@
...
+#   PT setup
+
+T_min = 22.5
+T_max = 27.5
+
+# temperature schedule: should take index (i) and number of processes (N) and return temperature (T)
+T_sched(i, N) = (10 .^(range(log10(T_min), log10(T_max), length=N)) )[i]
+
+# make replica for PT
+replica = Replica(MetropolisSampler(sys, 1.0, 1))
```
ColeMiles, line 33: Out of curiosity -- what happens if the user script makes two `Replica`'s? Maybe it should error, if that's detect-able.

ColeMiles (commented):
Seems good! There two lines that have to be changed pre-merge (related to the recent addition of `Requires`), and a few comments I have that could be implemented if you agree with me.

It might make sense to postpone the arbitrary-measurements to another pull request though. Then, you and other users of this can experiment with what works best for you.

src/ParallelTempering.jl
```diff
@@ -0,0 +1,168 @@
+using MPI
```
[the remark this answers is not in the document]
kbarros: Done

examples/PT_afm_heisenberg.jl
kbarros: Done.

src/ParallelTempering.jl
kbarros: Maybe we should have a longer discussion on Zoom with Matt about various design choices? I guess it can get improved in a follow-on pull request.

src/ParallelTempering.jl
kbarros: This seems like a good idea.

src/ParallelTempering.jl
mswwilson: The serialization sounds like a good idea. I'm open to changing the IO however makes the most sense, but had them all writing their own files at first just for my convenience. I'm open to discuss this or other ideas too

src/ParallelTempering.jl
mswwilson: Yeah, this would be nice to have some way to store arbitrary observables in each replica, like the dict you mention.

src/ParallelTempering.jl
mswwilson: Thanks for the nice suggestion - this makes a lot more sense! I've changed it to this now

examples/PT_afm_heisenberg.jl
mswwilson: It still seems to work fine if this happens. I tried it and realized that the spin configurations become de-synchronized with the  local running E, M in the Metropolis sampler if run! is called for each replica. I've added a line to fix this by resetting these quantities both at the start of each run, and after replica exchanges.

PR #8, opened 2021-12-03
Implementation of arbitrary spin-magnitudes and g-tensors

ColeMiles:
Implements specifying arbitrary spin magnitudes and g-tensors by changing `SpinSystem` to optionally receive a `Vector{SiteInfo}`, where `SiteInfo` is a new struct specifying the spin magnitude and g-tensor for a single site.

At `SpinSystem` construction time, all `SiteInfo` are propagated to symmetry-equivalent sites in the crystal, and all interactions are effectively rescaled to account for this info. Internally, `SpinSystem` continues to store an array of unit vectors, with all `SiteInfo` accounted for by rescaling interactions.


examples/reproduce_testcases.jl
```diff
@@ -49,31 +49,40 @@ function test_diamond_heisenberg_sf()
...
+end
+
 """
-    plot_S_cut(S, qz; max_ω, chop_ω)
+    plot_S_cut_afmdiamond(S, qz; max_ω, chop_ω)
 
 Plots cuts through the structure factor:
     (0, 0, qz) -> (π, 0, qz) -> (π, π, qz) -> (0, 0, 0)
-where qz is (c⃰ * qz). Only works on cubic lattices. To get unitful
+where qz is (c⃰ * qz). Assumes that S has 1BZ along (a, b) and
```
kbarros, line 74: Just now noticing the notation `c⃰`. Is this standard (asterisk directly above c?) If not, maybe overbar?

src/Ewald.jl
```diff
@@ -408,17 +408,29 @@ struct DipoleRealCPU <: InteractionCPU
...
     lattice = Lattice(crystal, latsize)
-    return DipoleRealCPU(strength .* precompute_dipole_ewald(lattice; extent=extent, η=η))
+    A = strength .* precompute_dipole_ewald(lattice; extent=extent, η=η)
+    # Conjugate each matrix by the correct g matrices
+    for b1 in 1:nbasis(crystal)
+        S1, g1 = sites_info[b1].S, sites_info[b1].g
+        for b2 in 1:nbasis(crystal)
+            S2, g2 = sites_info[b2].S, sites_info[b2].g
+            for ijk in CartesianIndices(axes(A)[3:end])
+                A[b1, b2, ijk] = g1' * S1 * A[b1, b2, ijk] * S2 * g2
```
kbarros, line 421: This looks good. I guess the magnetic moment is:
`m = mu_B * S * g`,

with the `g` tensor going to the _right_ of `S`, and `mu_B` the Bohr magneton. In this case, the Bohr magneton can be effectively absorbed into `A`?

src/Interactions.jl
```diff
@@ -198,30 +214,34 @@ end
...
-    -∑_i 𝐁 ⋅ 𝐦_i.
+    -∑_i 𝐁 ⋅ 𝐌_i.
 ```
 
-The magnetic moments are ``𝐦_i = μ_B g 𝐬_i`` where ``g`` is the g-factor or
-g-tensor, and the spin magnitude ``|𝐬_i|`` is typically a multiple of 1/2. The
+The magnetic moments are ``𝐌_i = μ_B g 𝐒_i`` where ``g`` is the g-factor or
+g-tensor, and the spin magnitude ``|𝐒_i|`` is typically a multiple of 1/2. The
 Bohr magneton ``μ_B`` is a physical constant, with numerical value determined by
 the unit system.
```
kbarros, line 223: In the code above (line 421 in Ewald.jl), it looked like you were using `m ~ S g` and not `m ~ g S`. Should we update this comment?

Do we handle the Bohr magneton yet in the code?

src/Interactions.jl
```diff
@@ -232,20 +252,36 @@ while `η` controls the direct/reciprocal-space tradeoff in the Ewald summation.
...
+=#
 
-function energy(spins::Array{Vec3}, field::ExternalField)
-    B = field.B
+struct ExternalFieldCPU
+    effBs :: Vector{Vec3}  # |S_b|gᵀB for each basis index b
+end
+
+function ExternalFieldCPU(ext_field::ExternalField, sites_info::Vector{SiteInfo})
+    effBs = [site.g' * site.S * ext_field.B for site in sites_info]
```
kbarros, line 264: This is where Bohr magneton should appear?

Also, I guess here we are defining the magnetic moment as `M ~ (S' g)' = g' S`, which explains the transpose. But we should figure out a consistent convention. Note that `g` could possibly (but rarely) have an antisymmetric part, so we need sure to be consistent with the `g` vs `g'` distinction.

src/PairInteractions.jl
```diff
@@ -130,78 +137,78 @@ function convert_quadratic(int::QuadraticInteraction{D}, cryst; tol=1e-6) where
...
     for (i, bonds) in enumerate(culled_bonds)
         for bond in bonds
             @unpack j, n = bond
             for cell in CartesianIndices(latsize)
-                Sᵢ = spins[i, cell]
-                Sⱼ = spins[j, offset(cell, n, latsize)]
-                E += Sᵢ ⋅ Sⱼ
+                sᵢ = spins[i, cell]
+                sⱼ = spins[j, offset(cell, n, latsize)]
+                E += sᵢ ⋅ sⱼ
```
kbarros, line 149: Question: It seems sometimes we use lower case s and sometimes upper case S. Is it intended that these have subtly different meanings?

src/StructureFactors.jl
```diff
@@ -145,43 +145,60 @@ function StructureFactor(snaps::Vector{Array{Vec3, 4}}, crystal; kwargs...)
...
+Updates `M` in-place to hold the magnetization vectors obtained by scaling `s`
+ by the appropriate spin magnitudes and g-tensors in `sites_info`.
+This function assumes `M` has a first index of length 3, which correspond
+ to the magnetization components. (Rather than storing an Array{Vec3}).
+"""
+function _compute_mag!(M, sys::SpinSystem)
+    for b in 1:nbasis(sys)
+        Sg = sys.sites_info[b].S * sys.sites_info[b].g
+        for idx in eachcellindex(sys)
+            M[:, b, idx] .= Sg * sys[b, idx]
```
kbarros, line 158: Bohr magneton? Also I realize now that `S g` and `g' S` have different shapes (1x3 vs 3x1 matrices). It is probably best to stick with 3x1 matrices if we want to interpret `M` as a 3-vector.

src/Symmetry/Crystal.jl
```diff
@@ -102,6 +102,13 @@ cell_volume(cryst::Crystal) = abs(det(cryst.lat_vecs))
 # This will remain 3 for the foreseeable future.
 dimension(cryst::Crystal) = 3
 
+"""
+    equiv_sites(crystal::Crystal, b::Int)
+
+Returns a list of all basis indices in the same symmetry equivalency class
+as the provided index `b`.
+"""
+equiv_sites(cryst::Crystal, b::Int) = filter(i->cryst.classes[i]==cryst.classes[b], 1:nbasis(cryst))
```
kbarros, line 111: Probably better to replace `1:nbasis(cryst)` with `eachindex(cryst.classes)`.

But now I am wondering if it's even better to use `findall`. Does this work?

`findall(==(cryst.classes[b]), cryst.classes)`

src/Symmetry/Crystal.jl
```diff
@@ -506,3 +513,81 @@ function validate(cryst::Crystal)
...
+    lat_vecs = lattice_vectors(a, a, a, 90, 90, 90)
+    basis_vecs = [
+        [0, 0, 0]/4,
+        [2, 2, 0]/4,
+        [1, 1, 1]/4,
+        [3, 3, 1]/4,
+        [2, 0, 2]/4,
+        [0, 2, 2]/4,
+        [3, 1, 3]/4,
+        [1, 3, 3]/4,
```
kbarros, line 566: I notice the ordering changed. Is this using the Sunny sorting now?

src/Systems.jl
```diff
@@ -70,22 +70,57 @@ end
...
+"""
+function propagate_site_info(crystal::Crystal, sites_info::Vector{SiteInfo})
+    # All sites not explicitly provided are by default S = 1, g = 1
+    all_sites_info = [SiteInfo(i, 1, 1) for i in 1:nbasis(crystal)]
+
+    specified_atoms = Int[]
+    for siteinfo in sites_info
+        @unpack site, S, g = siteinfo
+        (sym_bs, sym_gs) = all_symmetry_related_couplings(crystal, Bond(site, site, [0,0,0]), g)
+        for (sym_b, sym_g) in zip(sym_bs, sym_gs)
```
kbarros, line 87: Maybe here we need to be especially careful about the `g` vs `g'` convention?

src/Systems.jl
```diff
@@ -126,7 +161,12 @@ energy(sys::SpinSystem) = energy(sys.sites, sys.hamiltonian)
     field!(B::Array{Vec3}, sys::SpinSystem)
 
 Updates B in-place to contain the local field at each site in the
-system under `sys.hamiltonian`
+system under `sys.hamiltonian`. The "local field" is defined as
+
+``𝐁_i = -∇_{𝐬_i} ℋ / S_i``
+
+with ``𝐬_i`` the unit-vector variable at site i, and ``S_i`` is
+the magnitude of the associated spin.
```
kbarros, line 169: I guess this answers my question above.

docs/src/examples.md
```diff
@@ -70,39 +71,60 @@ crystal = Crystal(lat_vecs, [[0, 0, 0]], 227; setting="1")
...
+site (as they are all symmetry equivalent). We can also optionally omit this
+information, in which case all sites are assumed to be spin-1 with an identity
+``g``-tensor.
+
+Then, we will randomize the system so that all spins are randomly placed on the unit 
+sphere.
 
 ```julia
-sys = SpinSystem(crystal, interactions, (8, 8, 8))
+sites_info = [SiteInfo(1, 3/2, I(3))]
```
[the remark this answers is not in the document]
ColeMiles: That was what I had at first, but unfortunately that does not work out of the box:

```julia
julia> SiteInfo(1, 3/2, I)
ERROR: MethodError: Cannot `convert` an object of type UniformScaling{Bool} to an object of type StaticArrays.SMatrix{3, 3, Float64, 9}
Closest candidates are:
  convert(::Type{T}, ::Factorization) where T<:AbstractArray at /usr/share/julia/stdlib/v1.7/LinearAlgebra/src/factorization.jl:58
  convert(::Type{SA}, ::Tuple) where SA<:StaticArrays.StaticArray at ~/.julia/packages/StaticArrays/OWJK7/src/convert.jl:13
  convert(::Type{SA}, ::SA) where SA<:StaticArrays.StaticArray at ~/.julia/packages/StaticArrays/OWJK7/src/convert.jl:12
  ...
Stacktrace:
 [1] SiteInfo(site::Int64, S::Float64, g::UniformScaling{Bool})
   @ Sunny ~/Programming/Sunny.jl/src/Interactions.jl:44
 [2] top-level scope
   @ REPL[10]:1
```

`I(3)` is a `Bool` matrix, but the constructor for `SiteInfo` knows how to convert it to a `Mat3`.

docs/src/examples.md
```diff
@@ -70,39 +71,60 @@ crystal = Crystal(lat_vecs, [[0, 0, 0]], 227; setting="1")
...
 
 **(3)** Assembling a `SpinSystem` is straightforward -- provide the `Crystal`,
 the `Interaction`s, along with the dimensions of the simulation box in units of
-the lattice vectors. Then, we will randomize the system so that all spins are
-randomly placed on the unit sphere.
+the lattice vectors. We will also take this to be a spin-``3/2`` system with
+a identity ``g``-tensor, which we can specify by providing a `SiteInfo` on a single
+site (as they are all symmetry equivalent). We can also optionally omit this
+information, in which case all sites are assumed to be spin-1 with an identity
+``g``-tensor.
```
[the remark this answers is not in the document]
ColeMiles: Ah yeah, I see your point. I can make that change.

examples/reproduce_testcases.jl
```diff
@@ -13,30 +13,30 @@ Langevin dynamics.
...
-        heisenberg(J, Bond(3, 6, [0,0,0])),
+        heisenberg(J, Bond(1, 3, [0,0,0])),
     ]
-    sys = SpinSystem(crystal, interactions, (8, 8, 8))
+    sys = SpinSystem(crystal, interactions, (8, 8, 8), [SiteInfo(1, 3/2, I(3))])
     rand!(sys)
 
-    Δt = 0.02 / J       # Units of 1/K
-    kT = 4.             # Units of K
+    Δt = 0.02 / ((3/2)^2 * J) # Units of 1/K
```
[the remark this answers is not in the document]
ColeMiles: The motivation for writing it like this was because the example documentation I write:

"A rough rule of hand is that a reasonable value for integration timestep size is Δt≈0.02/(S^2∗J)"

and I just wanted to be explicit about how the Δt was being obtained.

examples/reproduce_testcases.jl
ColeMiles: The standard is an asterisk superscript, not directly over the c. I think I should just not try to use the unicode "asterisk above", and use "^*".

src/Ewald.jl
ColeMiles: Here, `S1` and `S2` are the scalar spin magnitudes, not the vector-spin values themselves. The idea is to absorb as much as possible into the dipole tensor.

You're right though that the Bohr magneton is still missing, and that it can also be absorbed into `A`. I'm planning to do this in a future "unit system" PR.

src/Interactions.jl
ColeMiles: Same as above -- that line was constructing the `A` tensor, and just trying to absorb as many factors as possible into it. Bohr magneton to come soon.

src/Interactions.jl
ColeMiles: Yeah, a Bohr magneton needs to appear here and in the construction of `A`.

The transpose here is because we want (ignoring the magneton): 

` E = -∑_i 𝐁 g 𝐒_i`

and I instead precompute an effective `effB = g' * S * 𝐁` so that during the simulations I can just compute:

`E = -∑_i effB ⋅ 𝐬_i`

where `𝐬_i` are the unit vectors for each spin.

src/PairInteractions.jl
ColeMiles: Yeah, the convention I'm working with is that lower-case s are the _unit vectors_ we internally store. Upper case S in the code should be a scalar spin magnitude. In the math within comments/documentation, I try to use this convention:

```𝐒_i = S_i 𝐬_i```

So, upper-case bold S is the full vector, upper-case S is the scalar spin magnitude, lower-case is the unit vector.

File: src/StructureFactors.jl (hunk 6 above)
Comment (ColeMiles) [reply src/StructureFactors.jl:158]: Same as above comments -- here `Sg` is the scalar spin magnitude times the g-tensor, so still a 3x3 matrix.

File: src/Symmetry/Crystal.jl (hunk 8 above)
Comment (ColeMiles) [reply src/Symmetry/Crystal.jl:566]: Yeah, we had previously discussed this briefly. There was some mixup between the vectors chosen by us manually, and by Spglib when we ask to construct all sites from the symmetry operations. I just changed these to agree with Spglib's list.

File: src/Systems.jl (hunk 9 above)
Comment (ColeMiles) [reply src/Systems.jl:87]: Yeah, this is a place to be careful. The convention I'm using everywhere is

`m_i = g s_i`

where `g` is applied to the left. Under this convention, I believe this is the correct way to transform since exchange matrices also are applied on the left.

File: src/Symmetry/Crystal.jl (hunk 7 above)
Comment (ColeMiles) [reply src/Symmetry/Crystal.jl:111]: I had forgetten about `findall`! That does work, and I agree it reads better. Will make the swap.

File: docs/src/examples.md [hunk 14]
```diff
@@ -70,39 +71,60 @@ crystal = Crystal(lat_vecs, [[0, 0, 0]], 227; setting="1")
...
+``Δt ≈ 0.02 / (S^2 * J)``.
 
 ```julia
-Δt = 0.02 / J       # Units of 1/K
-kT = 4.             # Units of K
+Δt = 0.02 / ((3/2)^2 * J) # Units of 1/K
+kT = 4.                   # Units of K
 α  = 0.1
-kB = 8.61733e-5     # Units of eV/K
+kB = 8.61733e-5           # Units of eV/K
```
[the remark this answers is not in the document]
ColeMiles: Yeah, the only place `kB` is used is down at the bottom of the example, in the plotting code so that I can plot frequency in terms of meV rather than K. So, I will move this line down to there.

src/StructureFactors.jl
ColeMiles: I will change the names / multiplication order here, actually, just to reflect that the convention is indeed m = g S. That will prevent confusion in the future, thanks for catching this.

docs/src/examples.md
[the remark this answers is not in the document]
kbarros: I was a bit puzzled by this, so I digged in. Like you say, this works:

```
julia> convert(Mat3, I(3))
3×3 SMatrix{3, 3, Float64, 9} with indices SOneTo(3)×SOneTo(3):
 1.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  1.0
```

But this errors
```julia
julia> convert(Mat3, I)
ERROR: MethodError: Cannot `convert` an object of type UniformScaling{Bool} to an object of type SMatrix{3, 3, Float64, 9}
Closest candidates are:
  convert(::Type{T}, ::Factorization) where T<:AbstractArray at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.7/LinearAlgebra/src/factorization.jl:58
  convert(::Type{SA}, ::Tuple) where SA<:StaticArray at /Users/kbarros/.julia/packages/StaticArrays/OWJK7/src/convert.jl:13
  convert(::Type{SA}, ::SA) where SA<:StaticArray at /Users/kbarros/.julia/packages/StaticArrays/OWJK7/src/convert.jl:12
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[11]:1
```

Why does this first one work?
```

julia> @which convert(Mat3, I(3))
convert(::Type{SA}, a::AbstractArray) where SA<:StaticArray in StaticArrays at /Users/kbarros/.julia/packages/StaticArrays/OWJK7/src/convert.jl:28
```
which has signature
```
@propagate_inbounds function convert(::Type{SA}, a::AbstractArray) where {SA <: StaticArray}
```

Interesting, `I(3)` but not `I` is being treated as an AbstractArray, which can be directly verified:

```
julia> I::AbstractArray
ERROR: TypeError: in typeassert, expected AbstractArray, got a value of type UniformScaling{Bool}
Stacktrace:
 [1] top-level scope
   @ REPL[13]:1

julia> I(3)::AbstractArray
3×3 Diagonal{Bool, Vector{Bool}}:
 1  ⋅  ⋅
 ⋅  1  ⋅
 ⋅  ⋅  1
```

It seems to me that StaticArrays needs to include another `convert` method for `UniformScaling`, but I guess that's not our problem for now.

examples/reproduce_testcases.jl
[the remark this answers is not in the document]
kbarros: Maybe do it in two steps then?
```
S = 3/2
...
Δt = 0.02 / (S^2 * J)
```

examples/reproduce_testcases.jl
kbarros: Sounds good. To be clear, the meaning is that `c^*` is the reciprocal lattice vector corresponding to lattice vector `c`?

src/Ewald.jl
kbarros: I see. Might help to write:
`A[b1, b2, ijk] = (S1*S2) * g1' * A[b1, b2, ijk] * g2`
which to my mind emphasizes "`S1*S2` is a scalar prefactor".

src/Interactions.jl
kbarros: Looks good. Minor comment, maybe write `E = -∑_i 𝐁^T g 𝐒_i` in the comments, i.e., with the tranpose on `B`. Then it makes sense to define `B_eff^T = 𝐁^T g` or equivalently `B_eff = g^T 𝐁`.

src/PairInteractions.jl
kbarros: This sounds good, except we should not try to use bolded `𝐒` as an actual Julia variable identifier (which you don't, just commenting...)

examples/reproduce_testcases.jl
ColeMiles: Yup!

src/StructureFactors.jl
kbarros: Wait, I'm confused, the code as written `M[:, b, idx] .= Sg * sys[b, idx]` makes sense now, no need to change anything?

src/Symmetry/Crystal.jl
kbarros: Ah, right.

src/Systems.jl
kbarros: Nice

kbarros (changes requested):
I am happy with this PR, I guess there were only a couple minor aesthetic things or comments to add, at your discretion.

src/Ewald.jl
ColeMiles: My intuition for writing it like this is that it's as if I just pulled out the inside of the expression (in parentheses):

`𝐒_i^T g^T A g 𝐒_j` = 
`𝐬_i^T (S_i g^T A g S_j) 𝐬_j`

in which case I should really swap the g and the S on both sides of the current code to be clear.

Do you prefer this (newly corrected) pulled-out-internals version, or the prefactor version? I have a moderate preference for the as-is with the g and S swapped, but am fine with either.

src/StructureFactors.jl
ColeMiles: I would just swap the code/names to be `g S` rather than `S g`. Since `S` here is a scalar, both ways are doing the same thing, it's just that this way will emphasize the convention `M = g S`.

kbarros (approved):

PR #9, opened 2022-01-10
Add feedback optimization for PT temperatures

mswwilson:


kbarros (approved):
Thanks all!

ColeMiles:
> Matt's PT changes look good to me. Also Cole updated some things in the docs regarding the unit system, etc. I think we can accept both, but I'm curious how these two different things ended up in the same PR?

That's on me! I noticed that there were some issues in the docs from when I merged `units`, and wanted to fix them. Was worried that pushing another commit onto main might introduce merge conflicts and make this PR more complicated to finish, so I asked Matt if I could just append them onto this PR. In hindsight, though, I should not have added it into this PR since it's wholly unrelated. My mistake!

ColeMiles:
I wanted to take a quick look through if that's ok. I can't right now, but I can by the end of the day.

kbarros:
Of course -- handing off to you.

src/ParallelTempering.jl
```diff
@@ -126,13 +180,121 @@ function gather_print(replica::Replica, data, fname::String)
...
+        hcat([vec(spins[i,:,:,:,:]) for i in 1:3]...)
+    )
+
+    for r in eachrow(xyz)
+        @printf(output, "%f\t%f\t%f\t%f\t%f\t%f\n", r...)
+    end
+    println(output, "")
+
+    return nothing
+end
```
ColeMiles, line 201: I find the series of `hcat`'s a bit hard to follow, is this version fine with you? (I don't think the `collect`'s are necessary?)

```julia
function xyz_to_file(sys::SpinSystem{3}, output::IOStream)
    sites = reinterpret(reshape, Float64, sys.lattice)
    spins = reinterpret(reshape, Float64, sys.sites)
    xyz = vcat(sites, spins)
    xyz = reshape(xyz, 6, :)

    for c in eachcol(xyz)
        @printf(output, "%f\t%f\t%f\t%f\t%f\t%f\n", c...)
    end
    println(output, "")

    return nothing
end
```

src/ParallelTempering.jl
```diff
@@ -126,13 +180,121 @@ function gather_print(replica::Replica, data, fname::String)
...
-Run w/ the command "mpiexec -n [n_procs] julia --project [julia_script.jl]".
+function feedback_update!(replica::Replica, N_updates::Int64; w::Float64=0.0, dT′::Float64=1e-4)
+
+    T_new = zeros(replica.N_ranks)
+
+    N_labeled = (replica.N_up + replica.N_down)    
+    f = ((N_labeled > 0) ? replica.N_down/N_labeled : 0.0)
+    
+    # gather temperatures and replica flow 'f' from all replicas
+    gv = MPI.Gather([1.0/replica.sampler.β, f], 0, MPI.COMM_WORLD)
```
ColeMiles, line 218: Tiny nitpick, but maybe a better variable name than `gv`?

ColeMiles: I'm also realizing that `replica.sampler.β` will error with Langevin sampling too, and I've provided no way to get the temperature back out of a `LangevinSampler`! I can add another commit to this PR adding a new function like `get_temp` for all samplers that returns their temperature that we can call here instead -- I'll change this and other lines that grab the temperature in my commit.

So that we don't get conflicts, I'll wait until after you've made your pass?

src/ParallelTempering.jl
```diff
@@ -144,55 +306,189 @@ function run!(replica::Replica, T_sched::Function;
         end
     end
 
-    # set sampler's energy and magnetization values
+    # initialize energy and equilibrate replica to it's distribution
     replica.sampler.E = energy(replica.sampler.system)
     replica.sampler.M = sum(replica.sampler.system)
```
ColeMiles, line 311: See the other comment below -- please use `reset_running_energy!` and `reset_running_mag!`

src/ParallelTempering.jl
```diff
@@ -144,55 +306,189 @@ function run!(replica::Replica, T_sched::Function;
...
-    C = 0.0
+    C  = 0.0
 
-    N_rex = 0
-    N_measure = 0
+    system_size = length(replica.sampler.system)
+
+    # record timeseries for minimum energies 
+    Emin = typemax(Float64)
+    mcs_Emin = Vector{Vector{Float64}}()
```
ColeMiles, line 447: This might be a micro-optimization that doesn't really matter, but maybe consider making this a `Vector{Tuple{Float64, Float64}}` (or `Vector{Tuple{Int, Float64}}`) and changing the line that pushes to it. Since the Tuple is constant size, I think this will avoid a ton of allocations.

src/ParallelTempering.jl
```diff
@@ -203,21 +499,27 @@ function run!(replica::Replica, T_sched::Function;
...
     )
 
-    # gather and print: rank, "down" exch. accepts, "up" exch. accepts, acceptance ratio   
+    # gather and print thermodynamic measurements
     gather_print(
         replica, 
-        [replica.rank, rex_accepts[1], rex_accepts[2], sum(rex_accepts)/N_rex], 
-        "replica_exchanges.dat"
+        [1.0/replica.sampler.β, M/system_size, X/system_size, U/system_size, C/system_size], 
+        "measurements.dat"
```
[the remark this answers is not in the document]
ColeMiles: Hm, I'm now realizing there are a lot of output files and maybe it doesn't make sense to make the names of *all* of them controllable. Not sure anymore!

src/ParallelTempering.jl
```diff
@@ -90,16 +104,56 @@ function replica_exchange!(replica::Replica)
...
-                  replica.sampler.system.sites, rex_rank, 3, MPI.COMM_WORLD)
-
+    MPI.Sendrecv!(        
+        deepcopy(replica.sampler.system.sites), rex_rank, 3,
+                 replica.sampler.system.sites , rex_rank, 3, 
+        MPI.COMM_WORLD
+    )
     # recalculate energy and magnetization of new state
     replica.sampler.E = energy(replica.sampler.system)
     replica.sampler.M = sum(replica.sampler.system)
```
ColeMiles, line 114: See comments below (noticed this one late). Please use `reset_running_energy!` and `reset_running_mag!`.

ColeMiles (changes requested):
I left some suggestions on various changes, but it looks awesome!

In reviewing your code, I noticed some areas where I didn't fully make the `LangevinSampler` expose the same API as the Metropolis-based samplers which you depend on in some functions. After you make your pass, I'll fix this so that everything should work with both samplers and add a commit to the PR if that's fine with you!

examples/FBOPT_2D_Ising.jl
```diff
@@ -0,0 +1,54 @@
...
+    print_interval=20
+)
+
+# optimized temperature schedule
+T_sched(i, N) = T_opt[i]
+
+# run parallel tempering
+run_PT!(
+    replica, 
+    T_sched;
```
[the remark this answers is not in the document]
mswwilson: Sure, I can do that

examples/PT_afm_diamond.jl
```diff
@@ -7,32 +7,43 @@ using MPI
...
-    heisenberg(J, Bond{3}(3, 6, [0,0,0])),
+    heisenberg(J, Bond{3}(1, 3, [0,0,0])),
 ]
 
-# spin system
+# spin system -- setup to use K
+S = 1 
 extent = (4,4,4)
-sys = SpinSystem(crystal, interactions, extent)
+sys = SpinSystem(crystal, interactions, extent, [SiteInfo(1, S, 1)]; μB=0.67171381563034223582, μ0=17.3497470317891588)
```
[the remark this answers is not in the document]
mswwilson: oops. It is in K, I just forgot to change the comment back..

src/ParallelTempering.jl
```diff
@@ -90,16 +104,56 @@ function replica_exchange!(replica::Replica)
     end
 
     # accept replica exch.: swap configurations
-    MPI.Sendrecv!(replica.sampler.system.sites, rex_rank, 3,
-                  replica.sampler.system.sites, rex_rank, 3, MPI.COMM_WORLD)
-
+    MPI.Sendrecv!(        
+        deepcopy(replica.sampler.system.sites), rex_rank, 3,
```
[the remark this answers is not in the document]
mswwilson: Before, I was incorrectly using Sendrecv!() like Sendrecv_replace(), which created a race condition. I needed the replica.sampler.system.sites to send before it receives that same quantity from its partner rank.

src/ParallelTempering.jl
```diff
@@ -144,55 +306,189 @@ function run!(replica::Replica, T_sched::Function;
...
+    # set replica neighbor β's  
+    for i in 1:2
+        if replica.nn_ranks[i] != -1
+            replica.nn_βs[i] = 1.0 / T_sched(replica.nn_ranks[i]+1, replica.N_ranks)
+        end
+    end
+
+    # initialize energy and equilibrate replica to it's distribution
+    replica.sampler.E = energy(replica.sampler.system)
+    replica.sampler.M = sum(replica.sampler.system)
```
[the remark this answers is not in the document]
mswwilson: I'll switch these over to use your interface. Also, I will make that change and add Metropolis.jl

src/ParallelTempering.jl
```diff
@@ -126,13 +180,121 @@ function gather_print(replica::Replica, data, fname::String)
...
+end
+
+"""
+Run a feedback-optimized parallel tempering simulation and return an optimized
+temperature set. No measurements are made, but this is used to diagnose replica
+flow in PT simulations. 
 """
-function run!(replica::Replica, T_sched::Function;
-                therm_mcs=1000, measure_interval=10, rex_interval=100, max_mcs=500_000, bin_size=1.0, print_hist=false)
+function run_FBOPT!(
```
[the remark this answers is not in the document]
mswwilson: Good point - will do

src/ParallelTempering.jl
mswwilson: This is much better, thanks.

src/ParallelTempering.jl
mswwilson: sounds good

src/ParallelTempering.jl
```diff
@@ -542,3 +614,23 @@ function run_PT!(
...
+    print_hist::Bool=false, 
+    print_xyz_ranks::Vector{Int64}=Int64[]
+)
+    kT_func = (i, N)->kT_sched[i]
+
+    return run_PT!(
+        replica, kT_func; 
+        therm_mcs, measure_interval, rex_interval, max_mcs, bin_size, print_hist, print_xyz_ranks
+    )
+end
```
ColeMiles, line 636: This syntax actually does work for passing the kwargs (and you won't have to update this function when the original method changes):

```julia
function run_PT!(replica::Replica, kT_sched::Vector{Float64}; kwargs...)
    kT_func = (i, N) -> kT_sched[i]
    return run_PT!(replica, kT_func; kwargs...)
end

```

ColeMiles (commented):
Awesome, thanks! Just one suggested change making the vector-accepting `run_PT!` a bit easier since you can pass variable number of keyword arguments through.

I can make those changes to `LangevinSampler` I alluded to, along with this other `run_PT!` change (if that seems fine with you), and push another commit onto here.

src/ParallelTempering.jl
mswwilson: Oh, gotcha. Good to know

mswwilson:
Sounds good. Thanks so much for the suggestions!

ColeMiles (approved):
Alright, I've added my commit which adds the `get_temp` function to all of the samplers and changed the PT code to use that function.

Please check that I haven't broken your PT examples in doing this! I don't currently have access to a cluster on which to test the PT code, and I'm having trouble running the scripts through MPI on my personal machine.

Other than testing I haven't broken anything, this seems good to go!

PR #10, opened 2022-01-26
Remove arbitrary-dimension lattices and systems

ColeMiles:
This PR removes the arbitrary-dimensionality functionality of many types throughout Sunny, as some components were effectively locking the dimensionality to 3 anyway. Additionally, 1D + 2D simulations can easily still be performed in the 3D framework, and nobody wants to do >3D simulations.

Effectively, three type parameters can be removed from many types throughout Sunny, and the dimensionality of many arrays is now fixed and statically known. This should buy us some budget for complexity when we inevitably need to add more type parameters for SU(N) simulations.

We also can remove `OnSiteQuadratic`, which only existed in the first place to dodge an awkward unknown-dimensionality problem we previously had.

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -180,7 +180,7 @@ meas_rate = 40     # Number of timesteps between snapshots of LLD to input to FF
                    # The maximum frequency we resolve is set by 2π/(meas_rate * Δt)
 dyn_meas = 400     # Total number of frequencies we'd like to resolve
 dynsf = dynamic_structure_factor(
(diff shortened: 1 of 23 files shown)


ColeMiles:
Note: There is some code throughout Sunny (particularly in `StructureFactors.jl`) which is written in a way such that it works for arbitrary-dimension input arrays. All of the code still works after this PR, but could maybe be re-written more simply since we know we're fixing 3D.

kbarros (approved):
LGTM, thanks Cole!

PR #11, opened 2022-01-28
Refactor bond-handling code in PairInteractions.jl into new type `BondTable`

ColeMiles:
In the previous code generating pair interactions, and using them to compute energies and fields, there was a lot of difficult-to-read code which mucked around with lists of bonds living on different sublattices.

This PR attempts to consolidate all of that logic into a new (internal) type `BondTable`, which exposes simple interfaces for iterating over all bonds, the culled bonds, or just those on a certain sublattice.

An additional goal of this refactor is that a single `Vector{Bond}` is more amenable to being put on the GPU than the `Vector{Vector{Bond}}` we worked with previously. The CUDA kernels I have been working on depend on this linear layout.

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -12,7 +12,7 @@ The high-level outline of performing a simulation is:
 1. Create a [`Crystal`](@ref), either by providing explicit geometry information
     (Example 1), or by loading a `.cif` file (Example 2).
 2. Using the `Crystal`, construct a collection of [Interactions](@ref).
-3. Assemble a [`SpinSystem`](@ref) using the newly created `Crystal` and Interactions, the size of the simulation box, and optionally information on the spin magnitude and ``g``-tensors of each site by passing a list of `SiteInfo`.
+3. Assemble a [`SpinSystem`](@ref) using the newly created `Crystal` and interactions, the size of the simulation box, and optionally information on the spin magnitude and ``g``-tensors of each site by passing a list of `SiteInfo`.
 4. Construct a sampler, either a [`LangevinSampler`](@ref) (Example 1), or a 
     [`MetropolisSampler`](@ref) (Example 2).
 5. Use the sampler directly to sample new states, or use it to perform [Structure factor calculations](@ref).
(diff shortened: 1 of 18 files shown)


docs/src/examples.md
```diff
@@ -12,7 +12,7 @@ The high-level outline of performing a simulation is:
 1. Create a [`Crystal`](@ref), either by providing explicit geometry information
     (Example 1), or by loading a `.cif` file (Example 2).
 2. Using the `Crystal`, construct a collection of [Interactions](@ref).
-3. Assemble a [`SpinSystem`](@ref) using the newly created `Crystal` and `Interaction`s, the size of the simulation box, and optionally information on the spin magnitude and ``g``-tensors of each site by passing a list of `SiteInfo`.
+3. Assemble a [`SpinSystem`](@ref) using the newly created `Crystal` and Interactions, the size of the simulation box, and optionally information on the spin magnitude and ``g``-tensors of each site by passing a list of `SiteInfo`.
```
kbarros, line 15: Should "Interaction" be lower case now?

docs/src/examples.md
```diff
@@ -12,7 +12,7 @@ The high-level outline of performing a simulation is:
 1. Create a [`Crystal`](@ref), either by providing explicit geometry information
     (Example 1), or by loading a `.cif` file (Example 2).
 2. Using the `Crystal`, construct a collection of [Interactions](@ref).
```
ddahlbom, line 14: Perhaps this should be lower case too.  I think the reference isn't working.  (Clicking the linking gives me 404.)

ddahlbom:
This all looks very clean and sensible to me. All the energy functions are easier to read (and presumably to construct) in this form, and it's nice to have the data packed together.

docs/src/examples.md
ColeMiles: Odd the reference appears to be working for me -- this one needs be capital for the `@ref` to work since it's referring to a section title.

ColeMiles:
> This all looks very clean and sensible to me. All the energy functions are easier to read (and presumably to construct) in this form, and it's nice to have the data packed together.

Thank you! Glad it seems reasonable.

I had to force-push this branch again following a rebase back onto main (due to the commit fixing plotting issues), but if all the tests pass I guess it's good to go?

kbarros:
I would personally say "go for it" if you think it's ready. I trust your judgment.

kbarros (approved):
Let's merge!

PR #13, opened 2022-02-14
Exchange MC

mswwilson:
PT/FBOPT and its examples are removed and replaced by REMC/FBO. 

Changes:
- User must initialize MPI in their script before instantiating replicas -- a function "init_MPI()" is provided for this
- Instead of tempering using kT, an arbitrary control parameter α is used 
- User must construct their own schedule for α before running REMC (using process rank)
- Control parameter schedule must start with smallest value as first entry (ex: largest kT <-> smallest β)

- User must define a function "set_α!(::Replica, ::Float64)" to assign updated α when using FBO
--- This "set_α!" should set a new SpinSystem in the sampler with updated Hamiltonian parameters, or can update 
--- parameters of the replica's AbstractSampler

- The integral in feedback updates has been debugged (to more reliably enforce normalization)
- After running FBO, the optimized α schedule is contained in the replicas and does not need to be set again
- Configurations are exchanged between replicas during exchange proposal step instead of after acceptance
- Net magnetization is measured as vector and histogram is added for magnetization norm.
- Replica flow, f(α), can be measured during REMC for diagnostic purposes


kbarros (approved):
Looks good to me. I like how the set_alpha!() function is user-specifiable and totally general.

Maybe for a future pull request you could allow parallelization over threads, without requiring MPI . I imagine that would be a common use case, and MPI is a somewhat heavy requirement.

Maybe hold off merging for a couple days in case @ColeMiles  wants to take a look?

kbarros:
Hi Matt, I think we're good to merge. Want to do the honors (push the button)?

PR #19, opened 2022-06-27
Replica exchage updates and general MC fixes to SU(N) Sunny

ddahlbom:


kbarros:
>  It now appears that all of Kipton's symmetry work is included in this as the final commit. I'm not sure if this is typical or if I messed something up.

For future reference, the solution was to check out this branch (mc_fixes), then rebase it on top of main, then `git push --force`. Now the history looks clean, and it's easier to see the actual diffs. Looks good to me.

PR #20, opened 2022-07-01
Offline viewers

kbarros:
Add a new command `offline_viewers()` to support offline Jupyter notebooks.


kbarros:
Matt, I put you as a reviewer in case you're curious about the Javascript hacks. This works for Jupyter notebooks, but it might not be possible for VSCode (unless we're willing to spend 0.5MB of RAM for each new viewer).

kbarros:
Also, we might want to think carefully about checking in the entirety of three.js -- it would seriously dilute our Julia/Javascript ratio. Maybe we can auto-download it at the time of Pkg install?

kbarros:
I guess it would be good to have input from @ColeMiles  and @ddahlbom  before landing this. Another possibility I was pondering is to make a separate package that carries just the three.js library, and then depend on that.

kbarros:
The Julia help desk pointed me here: https://pkgdocs.julialang.org/v1/artifacts/ This looks like the best way. I'll try to figure it out.

kbarros:
I'm merging this it fixes an important bug, and the changes are localized to SunnyGfx/

PR #28, opened 2022-08-20
Form Factor code with data table; CSV.jl dependency removed. 

sakibmatin:
J0 and J2 constants stored in Sunny/src/data/

Early draft. Any feedback on code or data table location with in Sunny?


kbarros:
Thanks for your pull request. Overall it looks good. I would suggest to remove the CSV.jl dependency because it brings many sub-dependencies, and we can probably do the parsing pretty easily ourselves.

Along these lines (but not directly related to this PR) is that another big offender is the CIF reading package. If we could one day write a custom parser instead, it may help with Sunny load times, although this should be tested. Julia 1.8 actually brings a nice macro to check load times,
https://julialang.org/blog/2022/08/julia-1.8-highlights/#package_load_timing
Would you be interested to try it with Sunny Sakib?

sakibmatin:
I agree. It should be straight foward to parse the data tables without CSV.jl, which clearly complicates the dependencies. I'll update the code.

I don't think I have the bandwidth to code up a custom cif parser for Sunny at this point. I'd be happy to try it out down the line.

kbarros:
Agreed. I was thinking the action item is to run the profiler and measure
what things are actually a bottleneck.

Kipton Barros, from phone

src/StructureFactors.jl
```diff
@@ -606,3 +606,72 @@ function accum_dipole_factor_wbasis!(res, S, lattice::Lattice)
...
+                b = csv_reader.b[k]
+                C = csv_reader.C[k]
+                c = csv_reader.c[k]
+                D = csv_reader.D[k]
+
+                if case == 0
+                    form1 = A*exp.(-a*(s.^2)) + B*exp.(-b*(s.^2)) + C*exp.(-c*(s.^2)) .+ D
+                elseif case == 2
+                    form2 = A*exp.(-a*(s.^2)) + B*exp.(-b*(s.^2)) + C*exp.(-c*(s.^2)) .+ D
+                    form_factor = form_factor = ((2.0-g)/g) .* (form2.*(s.^2) .+ form1)
```
kbarros, line 667: I assume this is a typo? `form_factor = form_factor = ...` ?

src/StructureFactors.jl
```diff
@@ -606,3 +606,72 @@ function accum_dipole_factor_wbasis!(res, S, lattice::Lattice)
         end
     end
 end
+
+
+
+""" FormFactor(q::Array{Float64,1}, elem::String)
+Magnetic Form-Factor with Gaussian Broadening. 
+Unitss for Q in inverse Angstrom, to match convention in data-table.
```
kbarros, line 614: Minor typo on `unitss`

src/StructureFactors.jl
```diff
@@ -606,3 +606,72 @@ function accum_dipole_factor_wbasis!(res, S, lattice::Lattice)
         end
     end
 end
+
+
+
+""" FormFactor(q::Array{Float64,1}, elem::String)
```
kbarros, line 612: Doc comment in triple quotes are space sensitive. Following Julia stdlib examples, I would suggest changes along these lines
```
"""    FormFactor(elem::String, q )

Compute the form factor for the wave vector `q` measured in inverse Angstroms, and associated with a magnetic ion of species `elem`.
[...]
"""
```

src/data/J0.dat
```diff
@@ -0,0 +1,97 @@
+Ion	A	a	B	b	C	c	D
```
kbarros, line 1: I think we move these two files to `Sunny.jl/assets/data/form_factor_J0.dat` and `form_factor_J2.dat`

src/StructureFactors.jl
```diff
@@ -606,3 +606,72 @@ function accum_dipole_factor_wbasis!(res, S, lattice::Lattice)
...
+
+    form1 = 0.0
+    form2 = 0.0
+    form_factor = 0.0
+
+    for case in 0:2:2
+        file_name = data_path*"J"*string(case)*".dat"
+        csv_reader = CSV.File(file_name)
+
+        for k in 1:length(csv_reader.Ion)
```
kbarros, line 653: Without CSV.jl, the code becomes even shorter:
```
for line in eachline(file_name)
    if line[begin] == '#'
        continue
    else
        (A, a, B, b, C, c, D) = split(line)
        ...
    end
end
```
I would recommend to put a `#` character on the first line to denote a comment.

src/StructureFactors.jl
```diff
@@ -606,3 +606,72 @@ function accum_dipole_factor_wbasis!(res, S, lattice::Lattice)
...
+    - Marshall W and Lovesey S WTheory of thermal neutron scattering Chapter 6 Oxford University Press (1971)
+    - Clementi E and Roetti C Atomic Data and Nuclear Data Tables14 pp 177-478 (1974)
+    - Freeman A J and Descleaux J P J. Magn. Mag. Mater. 12 pp 11-21 (1979)
+    - Descleaux J P and Freeman A J J. Magn. Mag. Mater. 8 pp 119-129 (1978) 
+"""
+function FormFactor(q::Array{Float64,1}, elem::String)
+    # relative path
+    data_path= string(@__DIR__)*"/data/"
+    
+    g = 5/6
```
kbarros, line 642: Where does g=5/6 come from? Note that Sunny already has a user-specifiable g-factor, so I want to make sure we're not conflicting with that.

src/StructureFactors.jl
sakibmatin: Yes. I'll fix the typo.

src/data/J0.dat
sakibmatin: Ok.

src/StructureFactors.jl
sakibmatin: Here, g is the Landé g-factor, which is different from the user defined g-factor/tensor in Sunny. 
The Landé g-factor only enters into the form factor calculation. 
g=6/5 is the place-holder for Er3 for testing. The final version of the code will have a dictionary for the Landé g-factors for different elements.

src/StructureFactors.jl
sakibmatin: Agreed.

kbarros:
This is looking good to me. Last thing I can see is to rename

`FormFactor() -> form_factor()`

Also, what's the intended usage? Will this be called by end users, or will this be built into the structure factor measurement?

Cc. @ddahlbom

ddahlbom:
It's great to see this. Experimentalists will want this off the bat.

My first instinct was to suggest this be applied behind the scenes as an option to `dynamic_structure_factor`, but I'm not opposed to making it publicly available. Perhaps a user would like to apply the form factor after calculating the structure factor so they can see the effect.

I think the main thing is the function should be adapted to act on arrays of `q` values, loading and parsing the text files only once. 

After than we can write an `apply_form_factor(sf::StructureFactor, elem::String, lande::Bool=false)` function that wraps this up. I'd be happy to do that -- I'm tweaking the structure factor stuff now -- but also happy if Sakib wants to give it a shot.

src/StructureFactors.jl
```diff
@@ -606,3 +606,99 @@ function accum_dipole_factor_wbasis!(res, S, lattice::Lattice)
...
+        "Co2"=>4.0/3.0,
+        "Ni2"=>5.0/4.0,
+        "Cu2"=>6.0/5.0,
+        "Zn2"=>0.0
+    )
+
+    g = 2.0 
+    if lande == true ; g = get(g_dict, elem, 2.0) ; end  
+    
+    s = q ./(4*π)
```
kbarros, line 680: Here `./` can be replaced with `/`

src/StructureFactors.jl
```diff
@@ -606,3 +606,99 @@ function accum_dipole_factor_wbasis!(res, S, lattice::Lattice)
...
+    
+    for case in 0:2:2
+        file_name = data_path*"form_factor_J"*string(case)*".dat"
+
+        for line in eachline(file_name)
+            (ion, A, a, B, b, C, c, D) = split(line)  
+            if ion == elem 
+                (A, a, B, b, C, c, D) = parse.(Float64, (A, a, B, b, C, c, D))
+                if case == 0
+                    form1 = A*exp.(-a*(s.^2)) + B*exp.(-b*(s.^2)) + C*exp.(-c*(s.^2)) .+ D
```
kbarros, line 692: Another option is:
```
form1 = @. A*exp(-a*(s^2)) + B*exp(-b*(s^2)) + C*exp(-c*(s^2)) + D
```

src/StructureFactors.jl
```diff
@@ -606,3 +606,99 @@ function accum_dipole_factor_wbasis!(res, S, lattice::Lattice)
...
+        "Ni2"=>5.0/4.0,
+        "Cu2"=>6.0/5.0,
+        "Zn2"=>0.0
+    )
+
+    g = 2.0 
+    if lande == true ; g = get(g_dict, elem, 2.0) ; end  
+    
+    s = q ./(4*π) 
+    form1 = 0.0 ; form2 = 0.0
```
kbarros, line 681: It looks like these two initializations are never used, and can be removed.

src/StructureFactors.jl
```diff
@@ -606,3 +606,99 @@ function accum_dipole_factor_wbasis!(res, S, lattice::Lattice)
...
+    form_factor = 0.0
+    
+    for case in 0:2:2
```
kbarros, line 684: Maybe better written:
```
for case in [0,2]
```

kbarros:
Sakib and I spent some time doing refactors today. David, if you're willing, please take over from here and integrate directly into the structure factor code. Sakib fork gives everybody write permissions, apparently.

Note that the is an optional "second order Landé correction" that is opt-in for the user. Not sure how you want to expose the user interface.

kbarros:
Something doesn't make sense to me in this table:
```
    g_dict = Dict{String,Float64}(
        "La3"=>0,
        "Ce3"=>6/7,
        "Pr3"=>4/5,
       ...
```
First, the ions La3 and Ce3 are not listed in the form factor data files. Should they be added? Second, a Landé g-factor of 0 seems to introduce a divide by zero. In that case, should elements like La3 be removed from the dict?

ddahlbom:
Sure, I'd be happy to integrate it when you guys are happy.

I already have some structure factor revisions in progress on another branch. Maybe it would be easiest if this PR is just merged, so the function and data are available in main. Then I can rebase on top of it and integrate it with the rest of the revisions. Does that make sense, or is another approach better?

sakibmatin:
> 

I agree. We should remove elements with a Landé g-factor of 0, as it introduces the division by zero error. 
We could keep "La3" and "Ce3" in the g_dict. We may choose to update the J0 and J2 constants later from other data sources if they become available.

kbarros:
Where did the table of Landé g factors come from? If "0" is a meaningful value, then maybe we should leave those entries, and add more checks to the subroutine.

kbarros:
The docs are unclear about what the input q is. I assume we will eventually want to get form factors for a `Vector{Vec3}`, i.e. a list of q-vectors? Does this function actually expect a `Vector{Float64}`, i.e. a list of magnitudes of q vectors?

ddahlbom:
For every point in momentum space we'll have a vector $\mathbf{q}$, which is just the coordinates in terms of the normalized reciprocal lattice vectors. The form factor calculation first involves calculating $\mathbf{q}' = \frac{\mathbf{q}}{4\pi}$ and then works with the scalar $\mathbf{q}' \cdot \mathbf{q}'$. We want to perform this operation for every point in momentum space.

So, ultimately we have to do this calculation on a list of 3-vectors.

kbarros:
This routine internally applies the $1/4\pi$ factor. I _think_ the input to this routine should be $q = |\mathbf{q}|$, from which we calculate $F(s)$ using $s = q / 2\pi$. From there, we can rescale the structure factor $S(\mathbf{q})$ according to $F(s)$. Does that sound right?

ddahlbom:
For what it's worth, this is how I calculated the qs in the past. I think Sunny has a built-in reciprocal lattice vector function, so writing the first function was probably unnecessary.
image

kbarros:
BTW, the relevant documentation is on this page: https://www.ill.eu/sites/ccsl/ffacts/ffactnode3.html

If you squint, this definition of $s = \sin \theta / \lambda$ can be interpreted as $s = 2 \pi q$, where $q$ denotes momentum transfer for a neutron of incoming momentum $2 \pi / \lambda$ In other words, $q = s / 2 \pi$. Not sure how I'm missing a factor of 2 with this calculation.

ddahlbom:
There was a factor of 2 I also found mysterious which Xiaojian insisted I put it...

kbarros:
I just force-pushed a commit which reintroduces the Lande factors with 0, because presumably these have some physical meaning (it might be a Dict we want to reuse in other contexts).

I have also revised the doc string, for future reference.

David and I had a long conservation and I think we agree with the input $q = |\mathbf q|$ should be a list of scalars representing the magnitudes of the $\mathbf q$-vectors.  I think we also agree that the `normed_qs()` subroutine is not needed.

PR #30, opened 2022-08-21
Graphics update

kbarros:
Cleaning up some of the SunnyGfx code. Now includes a more convenient way to open a browser. Would be nice to test on Linux and Windows before merging.

### DIFF
--- a/src/Sunny.jl
+++ b/src/Sunny.jl
@@ -101,7 +101,7 @@ include("WangLandau/WangLandau.jl")
 export WangLandau, spherical_cap_update, init_bounded!, run!
 
 include("SunnyGfx/SunnyGfx.jl")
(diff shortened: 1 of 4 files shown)


kbarros:
David pointed out that there is a problem on Linux: Any Linux application installed with Snap will not have access to files outside the user's home directory. This means that a standard Linux browser will typically not be able to open the temporary HTML files we create with `browser(viewer)`. The resolution here is to print a warning on Linux systems, and suggest that users include an optional argument ` browser(viewer; dir=...)`, where `dir` should be somewhere in the home directory, for accessibility to Snap applications. With this, `browser()` has been tested to work on both Linux and Mac. If we discover problems on Windows we can fix them in a follow up.

PR #33, opened 2022-09-05
Sf ff revisions

ddahlbom:
Main changes:
1. Structure factor interface changed slightly. For `dynamic_structure_factor`, user no longer sets `meas_rate`. Instead, they may set a `omega_max`, which will determine the down sampling. Number of frequencies resolved is specified with `num_omegas` (not `dyn_meas`). `dynΔt` is now just `dt`.
2. Form factor calculations have been revised. FF information is specified with `SiteInfo`. The corrections are precomputed prior to structure factor calculations and applied after calculating magnetization.
3. Some simple tests have been added for the structure factor material, but they are very preliminary: they don't test for correctness of results, just for errors. I have them commented out for now, but I run them before pushing. They would have caught the errors I introduced after doing the indexing refactor.
4. `sf_slice`, a function for cutting paths through a dynamical structure factor, now works on structure factors for which `reuduce_basis` and `dipole_factor` are false.

In general, this portion of the code is a little bit difficult to modify and expand because the structure factor data has a different number of indices depending on the stage of the calculation and the value of various optional parameters. At some point it would good to find some set of abstractions to make this a bit easier. Have various thoughts. To be continued.

### DIFF
--- a/examples/reproduce_testcases.jl
+++ b/examples/reproduce_testcases.jl
@@ -19,9 +19,7 @@ function test_diamond_heisenberg_sf()
     ]
     dims = (8, 8, 8)
     spin_rescaling = 3/2
-    ff_elem = "Fe2"
-    # ff_elem = nothing 
-    sys = SpinSystem(crystal, interactions, dims, [SiteInfo(1; spin_rescaling, ff_elem, ff_lande=5/2)])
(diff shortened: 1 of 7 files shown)


kbarros:
These changes look great to me. I cannot review in detail but I suspect any changes needed can be made as follow up PRs.

ddahlbom:
Ok, great. Thanks. I'll merge.

PR #39, opened 2022-09-23
Convert to TestItemRunner

kbarros:
Addresses #34

PR #40, opened 2022-09-24
Revamp docs

kbarros:
Big cleanup to the docs, which had fallen pretty stale.


kbarros:
This PR goes on top of #39 , so that one should probably be merged first.

PR #45, opened 2022-09-30
Remove uses of Logging macros

kbarros:
Note that `error()` throws an exception, whereas `@error` does not. Convert all instances to the former.

For consistency with the rest of the codebase, remove usages of `@warn`. Eventually we may want to replace all `println()` warning statements with `@warn`.

Additionally: Make system-wrapping interactions a warning rather than an error.


kbarros:
David gave thumbs up.

PR #48, opened 2022-10-18
Anisotropy refactor

kbarros:
Passes all existing tests, but probably more test cases should be added, especially to exercise the symmetry analysis associated with anisotropies.


src/Symmetry/LocalOperators.jl
```diff
@@ -0,0 +1,393 @@
...
+        # homogeneous polynomial of degree k
+        𝒪 = sum(t for t in 𝒪 if DP.degree(t) == k)
+        # Remaining coefficients must be real integers; make this explicit
+        𝒪 = DP.mapcoefficients(x -> Int(x), 𝒪)
+        return 𝒪
+    end
+end
+
+# Construct explicit N-dimensional marix representation of operator
+function operator_to_matrix(p; N)
```
ddahlbom, line 155: This is not important, but I wonder if this should be exported. Certainly it's nice to keep everything in terms of abstract symbols, but I can imagine wanting to see the concrete result.

src/Symmetry/LocalOperators.jl
```diff
@@ -0,0 +1,393 @@
...
+                (2/2)*(c6[5]*Jp¹-im*c6[9] *Jp¹)A[3] +
+                (1/2)*(c6[6]    -im*c6[8]     )A[2]
+
+    # Unpack gradient components
+
+    @label exit
+    dE_dx = +2real(dE_dJp)
+    dE_dy = -2imag(dE_dJp)
+    return (E, Vec3(dE_dx, dE_dy, dE_dz))
+end
```
ddahlbom, line 393: I have to admit that I think I would need a while to fully unpack everything here. But it's clearly very well written.

src/Symmetry/Printing.jl
```diff
@@ -192,13 +198,13 @@ end
...
-    print_bond_table(cryst::Crystal, max_dist)
+    print_symmetry_table(cryst::Crystal, max_dist)
 
 Pretty-prints a table of bonds, one for each symmetry equivalence class, up to a
 maximum bond length of `max_dist`. Equivalent to calling `print_bond(cryst, b)`
 for every bond `b` in `reference_bonds(cryst, max_dist)`.
 """
-function print_bond_table(cryst::Crystal, max_dist; digits=4, atol=1e-12)
+function print_symmetry_table(cryst::Crystal, max_dist; digits=4, atol=1e-12)
     for b in reference_bonds(cryst, max_dist)
```
ddahlbom, line 208: Took me a minute to realize information for a site is retrieved simply through `Bond(j,j,[0,0,0])`. But this is very nice.

src/Systems.jl
```diff
@@ -10,7 +10,7 @@ struct SpinSystem{N}
     _dipoles    :: Array{Vec3, 4}                   # Holds dipole moments: Axes are [Basis, CellA, CellB, CellC]
     _coherents  :: Array{CVec{N}, 4}                # Coherent states
     site_infos  :: Vector{SiteInfo}                 # Characterization of each basis site
-    S           :: Array{ComplexF64, 3}    
+    S           :: Array{ComplexF64, 3}             # TODO -- generate on demand instead
```
ddahlbom, line 13: I guess you found this broke something when removed?

src/Util.jl
```diff
@@ -38,65 +39,3 @@ end
     Ar = reinterpret(reshape, Float64, parent(A))
     return reshape(Ar, 3, 3, size(A)...)    # make sure this doesn't mess up indexing
 end
-
```
ddahlbom, line 41: Cleans things up to see these go. Going from \cdot to * for S \cdot B confused me at first, but this all makes sense.

test/test_shared.jl
```diff
@@ -6,16 +6,14 @@
 
 using Random
 using LinearAlgebra
+import WignerSymbols: clebschgordan, wigner3j
```
ddahlbom, line 9: Are these used?

test/test_symmetry.jl
```diff
@@ -213,63 +250,193 @@ end
...
+            randn(13)' * Sunny.stevens_operator_symbols[6]
+        (_, c2, _, c4, _, c6) = Sunny.operator_to_classical_stevens_coefficients(p, 1.0)
+
+        p_classical = Sunny.operator_to_classical_polynomial(p)
+        grad_p_classical = DP.differentiate(p_classical, Sunny.spin_classical_symbols)
+
+        E_ref = p_classical(Sunny.spin_classical_symbols => s)
+
+        gradE_ref = [g(Sunny.spin_classical_symbols => s) for g = grad_p_classical]
+        gradE_ref -= (gradE_ref⋅s)*s / (s⋅s) # Orthogonalize to s
```
ddahlbom, line 369: Recall discussion about this. Will probably need to ask you to explain more at a later point.

ddahlbom (approved):
Looks great to me!

test/test_shared.jl
ddahlbom: Sorry, I see where.

src/Symmetry/LocalOperators.jl
kbarros: Power users can always access this with `Sunny.operator_to_matrix()`, but I don't know if it's existence needs to be advertised to typical users. I could go either way though.

src/Symmetry/Printing.jl
kbarros: Thanks for catching this. I have updated the doc string and removed the `print_bond(cryst, site)` function with the new `print_site(cryst, site)` function. It is still true, however, that `Bond(j,j,[0,0,0])` is another way of specifying a site.

src/Util.jl
kbarros: With the move to abstract polynomials,  I observed that `S \cdot B ` became valid syntax without defining those `dot` overloads, and it's better to avoid "type piracy" where possible.

test/test_symmetry.jl
kbarros: We should have it documented. The question is: what should be the gradient of `X = |S|^2` with respect `S`? If you treat `X` as a scalar, the gradient is 0. If you treat `X` as a function of `S`, the gradient is `2S`. In practice, (e.g. for spin dynamics) we only use the part of the gradient that is _orthogonal_ to `S`, in which case it is always reasonable to treat `grad X = 0`. That makes it much easier to implement `energy_and_gradient_for_classical_anisotropy`.

kbarros: I added a comment to this effect in test_symmetry.jl

src/Systems.jl
kbarros: Good catch. This was easy to remove. See the most recent commit, [9341612](https://github.com/SunnySuite/Sunny.jl/pull/48/commits/93416129ff041daaefede434acd38c4fa661a0e9)

PR #50, opened 2022-10-26
SF intensity and sum rule revisions

ddahlbom:
This is a collection of minor revisions and bug fixes to make it easier to check sum rules. The intensities are also rescaled to match physical conventions. This makes it easier to compare results from Sunny directly with spin wave calculations.

Changes in behavior and features:

1. Accumulation of SF samples is now done with averaging, $\overline{x}_{n}= \overline{x}_{n-1}+\frac{x_{n} - \overline{x}_{n-1}}{n}$, so the intensity doesn't grow with the number of samples.
2. The Fourier transforms of the trajectories are now scaled by a factor of $\frac{1}{\sqrt{N_1N_2N_3}T}$, where $N_i$ is the dimension of the lattice along lattice vector $i$, and $T$ is the number of time steps. The resulting intensities are now comparable to what would be given by spin wave calculations. In particular, if you sum out out the $q$ and $\omega$ indices of $\textrm{Tr} \left( \mathcal{S} \left(\mathbf{q}, \omega \right) \right)$, the result will now be $N_1N_2N_3\vert S\vert^2$.
3. There is now an option to turn off application of the g-factor. (This is helpful to check the sum rule when the g-factor is anisotropic.)

Bug fixes:
1. I had introduced an indexing arithmetic error in one of the accumulation functions when I added offsets to the omega axis. Essentially all the negative omegas were off by a single index when `reduce_basis` was set to false. I never noticed because it is usually a subtle effect ($\Delta\omega$ is usually small) and negative energies are generally ignored. This manifested in small errors in the sum rule. This has been corrected. In the refactor, I'll add sum rule tests. We also won't be going back and forth between offset arrays and regular arrays, so I shouldn't be able to make the same mistake again.
2. I think there was an existing bug in `accum_dipole_factor_wbasis!`. Previously there `real(Sα[α, q_idx, :, :, :] * Sβ[β, q_idx, :, :, :])`, but I believe it should be `real(Sα[α, q_idx, :, :, :] * conj(Sβ[β, q_idx, :, :, :]))`. The `conj` has been added, and results are much more reasonable when `reduce_basis=false` and `dipole_factor=true`.
3. I added a hack to the `sf_slice` function so it doesn't throw an error when one of the spatial dimensions has size 1. This way users can use the function to get slices from effectively 2D systems. The fix is kind of ugly, but this function will not survive the upcoming refactor.

Testing:
I ran the code on two different models: one had eight sites per cell and was three dimensional, the other had one site per cell and was effectively 2D. I checked that the results were as expected for all permutations of `reduce_basis` and `dipole_factor` being true and false. I just the sum rule when `reduce_basis` was both true and false (but `dipole_factor` was always false, since it changes the intensities). I checked the slicing function on the 2D system.

### DIFF
--- a/src/StructureFactors.jl
+++ b/src/StructureFactors.jl
@@ -64,6 +64,7 @@ struct StructureFactor{A1, A2}
     num_omegas    :: Int                                     # Total number of snapshots to FT
     integrator    :: Union{SphericalMidpoint, SchrodingerMidpoint}
     plan          :: FFTW.cFFTWPlan{ComplexF64, -1, true, 6, NTuple{4, Int64}}
+    num_samples   :: Int64
 end
 
 Base.show(io::IO, sf::StructureFactor) = print(io, join(size(sf.sfactor), "x"),  " StructureFactor")
@@ -119,7 +120,7 @@ function StructureFactor(sys::SpinSystem{N}; bz_size=(1,1,1), reduce_basis=true,
 
     StructureFactor{typeof(sfactor), typeof(bz_buf)}(
         sfactor, spin_ft, ff_correction, bz_buf, sys.lattice, reduce_basis, dipole_factor,
-        bz_size, dt, meas_period, num_omegas, integrator, plan
+        bz_size, dt, meas_period, num_omegas, integrator, plan, 0
     )
 end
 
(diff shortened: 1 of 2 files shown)


src/StructureFactors.jl
```diff
@@ -427,8 +443,9 @@ function fft_spin_traj!(spin_traj::Array{ComplexF64};
                         plan::Union{Nothing, FFTW.cFFTWPlan}=nothing)
     if isnothing(plan)
         FFTW.fft!(spin_traj, (2,3,4,6))
+        spin_traj /= N
     else
-        spin_traj = plan * spin_traj
+        spin_traj = (plan * spin_traj) ./ 1e11
```
kbarros, line 448: 1e11 = inf?

kbarros (approved):
General comment -- does it make sense now to add unit tests, or should that wait for the big refactor?

src/StructureFactors.jl
ddahlbom: This is strange. I don't think this function is ever called. Only the one above it. I'll remove it.

ddahlbom: I guess it is called... I'll figure out what's going on.

src/StructureFactors.jl
```diff
@@ -407,6 +423,7 @@ of `spin_traj`.
 function fft_spin_traj!(res::Array{ComplexF64}, spin_traj::Array{Vec3};
                         plan::Union{Nothing, FFTW.cFFTWPlan}=nothing)
     @assert size(res) == tuplejoin(3, size(spin_traj)) "fft_spins size not compatible with spin_traj size"
+    println("here")
```
kbarros, line 426: I think this line accidentally made it through

ddahlbom: Sorry. Thanks. Removed it.

PR #52, opened 2022-11-15
Large N integrator fix

ddahlbom:
Removes the need to keep the spin matrices in memory for N > 5. Instead adds a function called `add_dipolar_field!(op::Array{ComplexF64, 2}, B::Vec3` which adds `B ⋅ S` to `op` by constructing the relevant matrix elements of the spin matrices on the fly. For N=6 this is marginally faster than keeping the matrices stored (3-5%). I also tried N=10, where it was about 10% faster than storing the matrices. So it seems like this is a reasonable approach to dealing with the N > 5 case, at least for now.

I added a test that ensures this function does what it should. I also added an `N=6` case to the spin rescaling tests, so that the `N > 5` case is at least run during routine testing.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ authors = ["Cole Miles <cmm572@cornell.edu>", "Kipton Barros <kbarros@lanl.gov>"
 version = "0.3.0"
 
 [deps]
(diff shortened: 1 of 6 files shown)


kbarros (approved):
Looks good to me!

PR #59, opened 2023-02-12
Inhomogeneous interactions

kbarros:
This PR introduces a new function `new_sys = to_inhomogeneous(sys)`. The new system `new_sys` is a copy of `sys` that allows inhomogeneous interactions. The new system supports functions like `set_exchange_at!` which sets an exchange interaction on just a single bond, ignoring symmetry and symmetry propagation.


kbarros:
At the types level, the main difference is that `System.interactions` now stores a union, which is either `Vector{Interactions}` or `Array{Interactions, 4}`. Which one determines the "mode" (homogeneous or inhomogeneous).
https://github.com/SunnySuite/Sunny.jl/blob/inhomogeneous/src/System/Types.jl#L56

There is now a function `is_homogeneous` to query this mode:
https://github.com/SunnySuite/Sunny.jl/blob/inhomogeneous/src/System/Interactions.jl#L20

Having more modes unfortunately introduces some code duplication. To minimize this, I tried to refactor various functions. For example, `energy` now calls `energy_aux`, which loops over contributions for a list of cells. In the case of inhomogeneous cells, the list of cells has only one cell at a time:
https://github.com/SunnySuite/Sunny.jl/blob/inhomogeneous/src/System/Interactions.jl#L204

PR #67, opened 2023-03-10
clean up outdated codes and add threaded PT

mswwilson:
Deleted old MC codes (will incrementally add back later) and add PT that works on single machine with threading. Can add example script in a later commit.


src/ParallelTempering.jl
```diff
@@ -0,0 +1,76 @@
...
+            PT.samplers[id₁].Δs, PT.samplers[id₂].Δs = PT.samplers[id₂].Δs, PT.samplers[id₁].Δs
+
+            PT.n_accept[id₁] += 1
+        end
+    end
+    PT.n_exch += 1
+end
+
+# run a parallel tempering simulation for 'nsweeps' MC sweeps
+function sample!(PT::ParallelTempering, n_sweeps::Int64, exch_interval::Int64)
```
kbarros, line 59: Could you please rename `n_sweeps` to `nsteps` ?

src/ParallelTempering.jl
```diff
@@ -0,0 +1,76 @@
...
+    system_ids::Vector{Int64}
+    # acceptance rate for replica exchanges btw (rank, rank+1)
+    n_accept::Vector{Int64}
+    # number of attempted exchanges
+    n_exch::Int64
+end
+
+function ParallelTempering(system::System, kT_sched::Vector{Float64}, propose)
+    n_replicas = length(kT_sched)
+    samplers = [LocalSampler(kT=kT, propose=propose) for kT in kT_sched]
```
kbarros, line 20: Minor point, but this could be written as `LocalSampler(; kT, propose)` for brevity. Let's do this for now.

Bigger point (for later): The user should have the flexibility to provide their own sampler. They might want Langevin instead of LocalSampler. Perhaps instead of `propose` we should ask the user to provide a sampler and an update function:
```
set_alpha!(sys::System, sampler, alpha::Float64)
```
where alpha will be between 0 and 1. This will support the replica exchange use case, etc.

src/ParallelTempering.jl
```diff
@@ -0,0 +1,76 @@
...
+        S₂′ = PT.samplers[id₁].ΔE / PT.samplers[id₂].kT
+        ΔS₂ = S₂ - S₂′
+
+        ln_P = ΔS₁ + ΔS₂
+
+        # acceptance criterion -- exchange labels and energy/magnetization for samplers
+        if ln_P >= 0 || rand(PT.systems[PT.system_ids[id₁]].rng) < exp(ln_P)
+            PT.system_ids[id₁], PT.system_ids[id₂] = PT.system_ids[id₂], PT.system_ids[id₁]
+
+            PT.samplers[id₁].ΔE, PT.samplers[id₂].ΔE = PT.samplers[id₂].ΔE, PT.samplers[id₁].ΔE
```
kbarros, line 49: The fields ΔE and Δs are only available for LocalSampler, but not for Langevin, for example.

Alternatively, I am thinking here we could call
```
system_alpha(...)
```
as defined above, for the swapped values of alpha.

kbarros:
Hi Matt, I put some comments thinking about how to generalize this to Langevin dynamics and Replica Exchange. Shall we merge this now and do those things as a second step?

PR #73, opened 2023-04-25
parallel tempering update

mswwilson:
-Generalize ParallelTempering.jl to accept LocalSampler or Langevin. 
-Add HistogramReweighting.jl for multiple histogram reweighting (Weighted Histogram Analysis Method) and calculation of ensemble averages from density of states. 
-Add BinnedArray for histogramming and measuring data in MC runs
-Move MC codes into a separate subdirectory 
-Remove outdated scripts from tests/mc_tests/

*Unit tests and example scripts will be added in subsequent commit
*A separate PR will be submitted for temperature selection that makes both parallel tempering and multiple histogram reweighting more robust


src/Sunny.jl
```diff
@@ -101,9 +98,18 @@ include("SunnyGfx/SunnyGfx.jl")
 include("SunnyGfx/CrystalViewer.jl")
 export view_crystal, offline_viewers, browser
 
-include("ParallelTempering.jl")
+include("MonteCarlo/Samplers.jl")
+export propose_uniform, propose_flip, propose_delta, @mix_proposals, LocalSampler
+
+include("MonteCarlo/BinnedArray.jl")
+export BinnedArray, reset!, get_keys, get_vals, get_pairs
```
kbarros, line 105: Consider removing these exports if they are not needed by users of Sunny (it creates a backward incompatibility danger).

src/Sunny.jl
```diff
@@ -101,9 +98,18 @@ include("SunnyGfx/SunnyGfx.jl")
...
+
+include("MonteCarlo/ParallelTempering.jl")
 export ParallelTempering, sample!
 
+include("MonteCarlo/HistogramReweighting.jl")
+export WHAM, ensemble_average
```
kbarros, line 111: Consider removing these exports also

src/Sunny.jl
```diff
@@ -101,9 +98,18 @@ include("SunnyGfx/SunnyGfx.jl")
 include("SunnyGfx/CrystalViewer.jl")
 export view_crystal, offline_viewers, browser
 
-include("ParallelTempering.jl")
+include("MonteCarlo/Samplers.jl")
```
kbarros, line 101: Can you put all the `include("MonteCarlo/...")` lines next to each other, following the style of previous imports?

src/MonteCarlo/ParallelTempering.jl
```diff
@@ -1,27 +1,42 @@
 # vanilla parallel tempering using Julia multithreading
-mutable struct ParallelTempering
+mutable struct ParallelTempering{N, SMP}
```
kbarros, line 2: Side-comment -- type parameters like {N} and {SMP} are useful for describing intent, and also useful for getting maximum performance, but may have the downside of creating more compilation work, and defeating precompilation. At some point we should go through all of Sunny and try to figure out whether it's possible to remove type parameters. For example, in this context, it may not be performance critical to have `N` statically known. To figure out it would require a lot of benchmarking.

kbarros (changes requested):

src/Sunny.jl
mswwilson: I was intending for these exports to be called by users so they can create histograms and collect statistics for measured observables with PT. 

I decided not to make a BinnedArray directly within PT so that the users are not forced to bin measurements. I may need to implement one more convenience function for using PT with Langevin so that the systems are returned in order of increasing temperature. Right now, you would need to do PT.systems[PT.system_ids] to get the correct order and then calculate measurements.

src/Sunny.jl
mswwilson: I also intended for WHAM() and ensemble_average() to be called by users themselves after running PT, as they can also choose to not use histogram reweighting. The ensemble_average function can be used for arbitrary observables that are defined by the user and collected with BinnedArrays as an average for each energy level, which are then reweighted to any desired temperatures. 

The ensemble_average function will also be useful for users to calculate results with g(E) from wang-landau and multicanonical simulations once I put those codes back in.

PR #78, opened 2023-05-27
Performance enhancements for inverse problem

Lazersmoke:
@kbarros , in anticipation of working on inverse problem (= approx 10^6 x forwards problem), I've made a few performance tweaks, mainly focused on making the memory allocations scale better (no GC anymore during long integrations 🎉).

These are mainly for my use but if they look useful enough for Sunny proper then I'm happy to make any style changes + squash the commits if needed.

Before:

benchPre

After:

benchPost

### DIFF
--- a/src/Integrators.jl
+++ b/src/Integrators.jl
@@ -94,6 +94,9 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
     (; Δt, atol) = integrator
 
     (B, s̄, ŝ, s̄′) = get_dipole_buffers(sys, 4)
+	
+	# Cache stores *norm* of (s̄ - s̄′), which is scalars
+    (progress,) = get_scalar_buffers(sys, 1)
     
     # Initial guess for midpoint
     @. s̄ = s
@@ -107,7 +110,9 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
         @. s̄′ = s + 0.5 * Δt * rhs_dipole(ŝ, B)
 
         # If converged, then we can return
-        if isapprox(s̄, s̄′, atol=atol* √length(s̄))
+		@. progress = norm(s̄ - s̄′)
+        if maximum(progress) <= atol
+        #if isapprox(s̄, s̄′, atol=atol* √length(s̄))
             # Normalization here should not be necessary in principle, but it
             # could be useful in practice for finite `atol`.
             @. s = normalize_dipole(2*s̄′ - s, sys.κs)
--- a/src/Integrators.jl
+++ b/src/Integrators.jl
@@ -94,9 +94,6 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
     (; Δt, atol) = integrator
 
     (B, s̄, ŝ, s̄′) = get_dipole_buffers(sys, 4)
-	
-	# Cache stores *norm* of (s̄ - s̄′), which is scalars
-    (progress,) = get_scalar_buffers(sys, 1)
     
     # Initial guess for midpoint
     @. s̄ = s
@@ -110,9 +107,7 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
         @. s̄′ = s + 0.5 * Δt * rhs_dipole(ŝ, B)
 
         # If converged, then we can return
-		@. progress = norm(s̄ - s̄′)
-        if maximum(progress) <= atol
-        #if isapprox(s̄, s̄′, atol=atol* √length(s̄))
+        if fast_isapprox(s̄, s̄′, atol* √length(s̄))
             # Normalization here should not be necessary in principle, but it
             # could be useful in practice for finite `atol`.
             @. s = normalize_dipole(2*s̄′ - s, sys.κs)
@@ -123,6 +118,19 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
     end
 
     error("Spherical midpoint method failed to converge to tolerance $atol after $max_iters iterations.")
+end
+
+function fast_isapprox(s̄, s̄′, atol)
+    sqTol = atol^2
+    acc = 0.
+    for i in eachindex(s̄)
+        diff = s̄[i] - s̄′[i]
+        acc += dot(diff,diff)
+        if acc > sqTol
+            return false
+        end
+    end
+    return true
 end
 
 
--- a/src/Integrators.jl
+++ b/src/Integrators.jl
@@ -120,7 +120,7 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
     error("Spherical midpoint method failed to converge to tolerance $atol after $max_iters iterations.")
 end
 
-function fast_isapprox(x, y; atol=0)
+function fast_isapprox(x, y; atol)
     sqTol = atol^2
     acc = 0.
     for i in eachindex(x)
--- a/src/Integrators.jl
+++ b/src/Integrators.jl
@@ -107,7 +107,7 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
         @. s̄′ = s + 0.5 * Δt * rhs_dipole(ŝ, B)
 
         # If converged, then we can return
-        if fast_isapprox(s̄, s̄′, atol* √length(s̄))
+        if fast_isapprox(s̄, s̄′,atol=atol* √length(s̄))
             # Normalization here should not be necessary in principle, but it
             # could be useful in practice for finite `atol`.
             @. s = normalize_dipole(2*s̄′ - s, sys.κs)
@@ -120,12 +120,12 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
     error("Spherical midpoint method failed to converge to tolerance $atol after $max_iters iterations.")
 end
 
-function fast_isapprox(s̄, s̄′, atol)
+function fast_isapprox(x, y; atol=0)
     sqTol = atol^2
     acc = 0.
-    for i in eachindex(s̄)
-        diff = s̄[i] - s̄′[i]
-        acc += dot(diff,diff)
+    for i in eachindex(x)
+        diff = x[i] - y[i]
+        acc += real(dot(diff,diff))
         if acc > sqTol
             return false
         end
@@ -257,7 +257,7 @@ function step!(sys::System{N}, integrator::ImplicitMidpoint; max_iters=100) wher
 
         @. Z″ = Z + ΔZ
 
-        if isapprox(Z′, Z″, atol=atol*√length(Z′))
+        if fast_isapprox(Z′, Z″, atol=atol*√length(Z′))
             @. Z = normalize_ket(Z″, sys.κs)
             @. sys.dipoles = expected_spin(Z)
             return
--- a/src/StructureFactors/BasisReduction.jl
+++ b/src/StructureFactors/BasisReduction.jl
@@ -6,7 +6,7 @@ function phase_averaged_elements(data, k::Vec3, sf::StructureFactor{N}, ffdata::
 
     for j in 1:NAtoms, i in 1:NAtoms
         phase = exp(im*(k ⋅ (rs[j] - rs[i])))
-        @. elems += phase * ffs[i] * ffs[j] * (@views data[:, i, j])
+        elems .+= phase .* ffs[i] .* ffs[j] .* view(data,:, i, j)
     end
 
     return SVector(elems)
--- a/src/StructureFactors/BasisReduction.jl
+++ b/src/StructureFactors/BasisReduction.jl
@@ -1,13 +1,13 @@
 function phase_averaged_elements(data, k::Vec3, sf::StructureFactor{N}, ffdata::Vector{FormFactor{FFType}}, ::Val{NCorr}, ::Val{NAtoms}) where {FFType, N, NCorr, NAtoms} 
-    elems = zero(SVector{NCorr, ComplexF64})
+    elems = zero(MVector{NCorr,ComplexF64})
     knorm = norm(k)
     ffs = ntuple(i -> compute_form(knorm, ffdata[i]), NAtoms)
     rs = ntuple(i -> sf.crystal.latvecs * sf.crystal.positions[i], NAtoms)
 
     for j in 1:NAtoms, i in 1:NAtoms
         phase = exp(im*(k ⋅ (rs[j] - rs[i])))
-        elems += phase * ffs[i] * ffs[j] * view(data, :, i, j)  # This view allocates
+        @. elems += phase * ffs[i] * ffs[j] * (@views data[:, i, j])
     end
 
-    return elems
+    return SVector(elems)
 end
--- a/src/StructureFactors/SampleGeneration.jl
+++ b/src/StructureFactors/SampleGeneration.jl
@@ -102,25 +102,22 @@ function no_processing(::StructureFactor)
     nothing
 end
 
-# ddtodo: Plan FFT
 function accum_sample!(sf::StructureFactor)
-    (; data, idxinfo, samplebuf, nsamples) = sf
+    (; data, idxinfo, samplebuf, copybuf, nsamples, fft!) = sf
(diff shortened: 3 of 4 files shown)


kbarros:
This looks very interesting! Would you mind attaching bench.jl to a comment in this PR? We can add a unit test that makes sense it remains GC free.

Lazersmoke:
Here is my `bench.jl`!

Note that it's only zero allocation per `step!` and fast for homogeneous Heisenberg interactions, currently. It may be low-hanging fruit to extend this to other interactions and the non-homogeneous case; all that needs to be done is replace all instances of `offsetc` by a corresponding `fastcircshift3!` version in `Interactions.jl`.

The unit test can probably just be `0 == @allocated step!(...)`; it probably doesn't need the full benchmarks as in `bench.jl`.

[bench.txt](https://github.com/SunnySuite/Sunny.jl/files/11583320/bench.txt)

src/StructureFactors/SampleGeneration.jl
```diff
@@ -113,12 +113,16 @@ function accum_sample!(sf::StructureFactor)
...
 
     # Transfer to final memory layout while accumulating new samplebuf
-    for ω in 1:nω, cell in CartesianIndices(latsize), j in 1:natoms, i in 1:natoms
+    cache = zeros(ComplexF64, size(data)[4:7])
+    for j in 1:natoms
+      for i in 1:natoms
         for (ci, c) in idxinfo 
-            α, β = ci.I
-            old = data[c,i,j,cell,ω]
-            data[c,i,j,cell,ω] = old + (samplebuf[α,cell,i,ω] * conj(samplebuf[β,cell,j,ω]) - old) / nsamples[1]
```
kbarros, line 120: There might be a semantic difference in the way that the running average had been calculated. By subtracting the mean (`old`) and then adding it back, it the previous version could possibly be reducing floating point round-off error. See here for an example: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online . It's not clear if this difference in floating point roundoff is important for this context.

src/StructureFactors/SampleGeneration.jl
```diff
@@ -113,12 +113,16 @@ function accum_sample!(sf::StructureFactor)
...
+          α, β = ci.I
+          @. cache = @views samplebuf[α,:,:,:,i,:] * conj(samplebuf[β,:,:,:,j,:]) - data[c,i,j,:,:,:,:]
```
kbarros: Is it necessary to have a cache here? Could the division by `nsamples[1]` included in a one-liner to update `data`?

src/Integrators.jl
```diff
@@ -107,7 +110,9 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
         @. s̄′ = s + 0.5 * Δt * rhs_dipole(ŝ, B)
 
         # If converged, then we can return
-        if isapprox(s̄, s̄′, atol=atol* √length(s̄))
+		@. progress = norm(s̄ - s̄′)
+        if maximum(progress) <= atol
```
kbarros, line 115: First, it is surprising to learn that `isapprox` is allocating -- in principle, this could be implemented in a non-allocating way. The semantics of the old convergence condition is (I think):
```julia
@. progress = norm(s̄ - s̄′)^2
cond = sum(progress .^2) <= atol* √length(s̄)
```
The new condition uses a maximum instead, which might or might not be what we want. 

I wonder if it's better to have a custom, non-allocating implementation of isapprox, and then use it uniformly in this file? Then the code would be easy to read and we wouldn't need the extra scalar buffers.

src/StructureFactors/StructureFactors.jl
```diff
@@ -96,8 +98,13 @@ function StructureFactor(sys::System{N}; Δt, nω, measperiod,
     na = natoms(sys.crystal)
     ncorr = length(correlations)
     samplebuf = zeros(ComplexF64, nops, sys.latsize..., na, nω) 
+    copybuf = zeros(ComplexF64, sys.latsize..., nω) 
     data = zeros(ComplexF64, ncorr, na, na, sys.latsize..., nω)
 
+    # Normalize FFT according to physical convention
+    normalizationFactor = 1/(nω * √(prod(sys.latsize)))
+    fft! = normalizationFactor * FFTW.plan_fft!(samplebuf, (2,3,4,6))
```
kbarros, line 106: Thanks for adding this!

src/Integrators.jl
```diff
@@ -157,11 +157,11 @@ end
 
 
 function rhs_langevin!(ΔZ::Array{CVec{N}, 4}, Z::Array{CVec{N}, 4}, ξ::Array{CVec{N}, 4},
-                        B::Array{Vec3, 4}, integrator::Langevin, sys::System{N}) where N
+                        B::Array{Vec3, 4}, circBuf, integrator::Langevin, sys::System{N}) where N
```
kbarros, line 160: For consistency, annotate type of `circBuf` here?

Also, Julia style avoids the use of camel-case `circBuf`. Another style option is snake-case, `long_variable_name`. In the case, let's use `circbuf` since it's easy to read.

src/System/Interactions.jl
```diff
@@ -282,7 +282,7 @@ end
 
 
 # Updates B in-place to hold negative energy gradient, -dE/ds, for each spin.
-function set_forces!(B, dipoles::Array{Vec3, 4}, sys::System{N}) where N
+function set_forces!(B, dipoles::Array{Vec3, 4}, buf::Array{Vec3,4}, sys::System{N}) where N
```
kbarros, line 285: Is this still `circbuf` ? If so, let's keep naming consistent.

src/System/Interactions.jl
```diff
@@ -372,6 +373,107 @@ function set_forces_aux!(B, dipoles::Array{Vec3, 4}, ints::Interactions, i::Int,
     end
 end
 
+function set_forces_aux_homog!(B, dipoles::Array{Vec3, 4}, buf::Array{Vec3,4}, ints::Interactions, i::Int, sys::System{N}) where N
```
kbarros, line 376: Big picture question -- is there a way to get the performance benefits while avoiding the code duplication? Perhaps we can discuss this on a Zoom call.

src/System/Interactions.jl
```diff
@@ -372,6 +373,107 @@ function set_forces_aux!(B, dipoles::Array{Vec3, 4}, ints::Interactions, i::Int,
...
+                B[cellᵢ, bond.i] -= J * 2sⱼ*(sᵢ⋅sⱼ)
+                B[cellⱼ, bond.j] -= J * 2sᵢ*(sᵢ⋅sⱼ)
+            elseif sys.mode == :SUN
+                error("Biquadratic currently unsupported in SU(N) mode.")
+            end
+        end
+    end
+end
+
+function fastcircshift3!(dest,src,shift)
```
kbarros, line 447: I am curious what happens if you try Julia's built-in `circshift!` -- does it not work well for arrays of 3-vectors?

src/System/System.jl
```diff
@@ -419,32 +419,32 @@ function polarize_spins!(sys::System{N}, dir) where N
...
-function get_scalar_buffers(sys::System{N}, numrequested) where N
+@inline function get_scalar_buffers(sys::System{N}, numrequested) where N
     numexisting = length(sys.scalar_buffers)
     if numexisting < numrequested
         for _ in 1:(numrequested-numexisting)
             push!(sys.scalar_buffers, zeros(Float64,size(sys.dipoles)))
         end
     end
-    return sys.scalar_buffers[1:numrequested]
+    return @view sys.scalar_buffers[1:numrequested]
```
kbarros, line 429: Style-wise, how about `view(sys.scalar_buffers, 1:numrequested)`

src/System/System.jl
```diff
@@ -419,32 +419,32 @@ function polarize_spins!(sys::System{N}, dir) where N
     end
 end
 
-function get_scalar_buffers(sys::System{N}, numrequested) where N
+@inline function get_scalar_buffers(sys::System{N}, numrequested) where N
```
kbarros, line 422: Is `@inline` actually having a benefit here? If not, I'd prefer to remove it.

kbarros (changes requested):

kbarros:
Most of these changes look great! I would like to spend more time with you thinking about whether it's possible to remove the code duplication in `set_forces_aux_homog!`, so perhaps we can split that commit into a separate PR. For the other parts, I think we can merge them more quickly after the comments are addressed. There are also a couple remaining minor questions about a non-allocating `isapprox` (should be doable in principle) and how to best implement the online running average. @ddahlbom is currently visiting LANL and helped me with the review.

src/System/Interactions.jl
Lazersmoke: Built-in `circshift!` is generic in the number of indices. For vectors and matrices it is non-allocating, but it is allocating (likely because it doesn't inline aggressively enough?) for rank 3 tensors and higher. `fastcircshift3!` is just exactly built-in `circshift!` specialized to rank 3 tensors and unrolled explicitly.

src/StructureFactors/SampleGeneration.jl
Lazersmoke: The semantics and algorithm are the same, it's just in two steps with explicit copy buffer instead of one step with implicit (allocated on demand) copy buffer.

To see that the copy buffer is required, note that the result of the first step (computing `x_n - \bar x_{n-1}`) must be stored somewhere and, at the same time, the original `\bar x_{n-1}` must be stored somewhere else.

The alternative to the copy buffer is to to use `samplebuf[α,:,:,:,i,:]` as the copy buffer, but this clobbers the `samplebuf`. But that might be OK, I'm not sure what Sunny's semantics/assumptions about that are!

src/System/Interactions.jl
Lazersmoke: I think so! This is (at least related to) the "possible low hanging fruit." We can discuss sometime this week after the Monday holiday

src/System/System.jl
Lazersmoke: Yes. The `view(sys.scalar_buffers,1:numrequested)` has to survive a function boundary if this isn't `@inline`, in which case it must be allocated.

src/StructureFactors/BasisReduction.jl
```diff
@@ -1,13 +1,13 @@
...
-    elems = zero(SVector{NCorr, ComplexF64})
+    elems = zero(MVector{NCorr,ComplexF64})
     knorm = norm(k)
     ffs = ntuple(i -> compute_form(knorm, ffdata[i]), NAtoms)
     rs = ntuple(i -> sf.crystal.latvecs * sf.crystal.positions[i], NAtoms)
 
     for j in 1:NAtoms, i in 1:NAtoms
         phase = exp(im*(k ⋅ (rs[j] - rs[i])))
-        elems += phase * ffs[i] * ffs[j] * view(data, :, i, j)  # This view allocates
+        @. elems += phase * ffs[i] * ffs[j] * (@views data[:, i, j])
```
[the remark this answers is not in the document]
Lazersmoke: Yep, just checked it's equivalent.

What's *not* equivalent is `@views` at the start of the line, which materializes the `ffs` and is slow, which is exactly the class of bug you avoid by using `view` over `@views` lol.

src/Integrators.jl
Lazersmoke: The difference is L2 vs Linfinity norm. Conceptually, the L2 version is like a soft-spin constraint and Linfinity is like a hard-spin constraint---it's a real physics question about which is a better stopping criterion for the method.

I think custom `isapprox` is a very good plan. Ideally, it doesn't even need to check every spin and probably could stop after the first one in the most common case.

src/StructureFactors/SampleGeneration.jl
Lazersmoke: Division is separated out here so there is one float divide and N float multiplys rather than N float divisions (minor speed boost). Could be `data[...] .+= copybuf .* (1/nsamples[1])` instead

kbarros: > Could be `data[...] .+= copybuf .* (1/nsamples[1])` instead

I like this version. Also worth checking whether the Julia compiler optimizes that down to `data[...] .+= copybuf ./ nsamples[1])`.

src/Integrators.jl
kbarros: > I think custom `isapprox` is a very good plan. Ideally, it doesn't even need to check every spin and probably could stop after the first one in the most common case.

Yes, the possibility of "early return" is an advantage of Linfinity, and I agree that it's not clear if there's a strong physics reason to prefer one norm over the other. If we define a custom `isapprox` then it would be easy to play with it later.

src/System/System.jl
kbarros: It seems you have done some benchmarks. Let's discuss this on our Zoom call.

src/StructureFactors/SampleGeneration.jl
kbarros: I verified your result. The compiler will not perform the optimization, even with `@fastmath` enabled.

```julia

function f(a, b)
    @fastmath a .= a ./ b
end

function g(a, b)
    a .= a .* (1/b)
end

@code_native f(a, b)
@code_native g(a, b)
```


```

**Comment by Lazersmoke** [comment]
Finally was able to convince github to accept my access token 🙏 

This should (hopefully) pass the refdata test now!

**Review by kbarros — CHANGES_REQUESTED** [review]

File: src/StructureFactors/BasisReduction.jl [hunk 12]
```diff
@@ -1,13 +1,13 @@
...
-    elems = zero(SVector{NCorr, ComplexF64})
+    elems = zero(MVector{NCorr,ComplexF64})
     knorm = norm(k)
     ffs = ntuple(i -> compute_form(knorm, ffdata[i]), NAtoms)
     rs = ntuple(i -> sf.crystal.latvecs * sf.crystal.positions[i], NAtoms)
 
     for j in 1:NAtoms, i in 1:NAtoms
         phase = exp(im*(k ⋅ (rs[j] - rs[i])))
-        elems += phase * ffs[i] * ffs[j] * view(data, :, i, j)  # This view allocates
+        elems .+= phase .* ffs[i] .* ffs[j] .* view(data,:, i, j)
```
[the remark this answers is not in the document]
Lazersmoke: Nope, it changes the meaning because it tries to do `view.(data,:,i,j)` which doesn't make sense (no method `view(::ComplexF64,::Colon,...)`)

So it can either be the `@.`, or use `view(...)` instead of `@view`, but not both.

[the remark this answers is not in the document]
Lazersmoke: Nope, it changes the meaning because it tries to do `view.(data,:,i,j)` which doesn't make sense (no method `view(::ComplexF64,::Colon,...)`)

So it can either be the `@.`, or use `view(...)` instead of `@view`, but not both.

src/Integrators.jl
```diff
@@ -120,6 +120,19 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
     error("Spherical midpoint method failed to converge to tolerance $atol after $max_iters iterations.")
 end
 
+function fast_isapprox(s̄, s̄′, atol)
```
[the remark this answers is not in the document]
Lazersmoke: It needs a `real(...)` because the dot product is complex and ruins the type of `acc`. But it gets optimized away in the real case so it's for free.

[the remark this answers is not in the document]
kbarros: If I understand your suggestion, inside the definition of `fast_isapprox` we would replace `dot(diff, diff) -> real(dot(diff, diff))`. This is the correct behavior.

src/StructureFactors/BasisReduction.jl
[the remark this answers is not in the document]
kbarros: I see. Let's go with the explicit `.` as you have written them.

kbarros:
Thanks @Lazersmoke for your hard work on this.

PR #84, opened 2023-06-14
Add `binned_histogram`

Lazersmoke:
This isn't finished/polished yet (opening this PR to track progress).

This is the alternative interface to `intensities` which is designed to more closely align with what Mantid/Horace output. In particular, the new type `BinningParameters` is supposed to be isomorphic to the data Mantid uses to describe an "MDHistoWorkspace".

Example usage:
```julia
# Histogram a 1D cut with from [0,1,0] to [1/2,0,0] (with transverse width 0.3)
params = Sunny.one_dimensional_cut_binning_parameters(sf,[0,1,0],[1/2,0,0],30,0.3)
display(params)

# sunHist: Histogramed intensity data
# sunCounts: Number of scattering vectors per bin (NOT uniform)
sunHist,sunCounts = Sunny.binned_histogram(sf,params,Sunny.DipoleFactor(sf),kT,formfactors)

# Axes labelled by bin centers
mx = Sunny.axes_bincenters(params)

fig3 = Plots.heatmap(mx[1],mx[4],sunCounts[:,1,1,:]')
logIntensity = log10.(sunHist[:,1,1,:]' ./ sunCounts[:,1,1,:]')
fig4 = Plots.heatmap(mx[1],mx[4],logIntensity;ylabel="E (meV)",color=cgrad(:viridis,scale = :lin,rev = false),xticks = ([mx[1][1],mx[1][end]], ["[0,1,0]","[1/2,0,0]"]))

display(Plots.plot(fig3,fig4;layout = @layout([a{0.2h};b])))
```
The resulting intensity histogram looks something like this:

image

### DIFF
--- a/src/StructureFactors/DataRetrieval.jl
+++ b/src/StructureFactors/DataRetrieval.jl
@@ -118,9 +118,8 @@ end
 
 # This places one histogram bin around each possible Sunny scattering vector.
 # This is the finest possible binning without creating bins with zero scattering vectors in them.
-function unit_resolution_binning_parameters(sf::StructureFactor) 
-    ωvals = ωs(sf)
-    numbins = (size(sf.samplebuf)[2:4]...,length(ωvals))
+function unit_resolution_binning_parameters(ωvals,latsize) 
+    numbins = (latsize...,length(ωvals))
     # Bin centers should be at Sunny scattering vectors
     maxQ = 1 .- (1 ./ numbins)
     total_size = (maxQ[1],maxQ[2],maxQ[3],maximum(ωvals)) .- (0.,0.,0.,minimum(ωvals))
@@ -129,6 +128,11 @@ function unit_resolution_binning_parameters(sf::StructureFactor)
     binend = (maxQ[1],maxQ[2],maxQ[3],maximum(ωvals)) .+ (binwidth ./ 2)
 
     return BinningParameters(binstart,binend,binwidth)
+end
+function unit_resolution_binning_parameters(sf::StructureFactor) 
+    ωvals = ωs(sf)
+    latsize = size(sf.samplebuf)[2:4]
+    return unit_resolution_binning_parameters(ωvals,latsize)
 end
 
 # Creates BinningParameters which implement a 1D cut in (Qx,Qy,Qz) space.
@@ -145,7 +149,7 @@ end
 #   2. Fist transverse Q direction
 #   3. Second transverse Q direction
 #   4. Energy
-function one_dimensional_cut_binning_parameters(sf::StructureFactor,cut_from_q,cut_to_q,cut_bins::Int64,cut_width;plane_normal = [0,0,1],cut_height = cut_width)
+function one_dimensional_cut_binning_parameters(ωvals::Vector{Float64},cut_from_q,cut_to_q,cut_bins::Int64,cut_width;plane_normal = [0,0,1],cut_height = cut_width)
     # This covector should measure progress along the cut in r.l.u.
     cut_covector = normalize(cut_to_q - cut_from_q)
     # These two covectors should be perpendicular to the cut, and to each other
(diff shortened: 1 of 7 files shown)


Lazersmoke:
TODO's on this include:
- Needs a separate function to treat powder averaging, since `BinningParameters` as written only makes sense for linear coordinates, and spheres are nonlinear
- Automatic conversion of Mantid/Horace output files (in HDF format) into `BinningParameters` objects. The manual conversion is not particularly difficult, but it is annoying to get right.
- Make it work with `SpinWaveTheory`. There are two possible ways to implement this, depending on whether we trust the spin wave theory calculation at `q` vectors which are not compatible with the periodic boundary conditions.

kbarros:
This sounds relevant to @ddahlbom

src/StructureFactors/DataRetrieval.jl
```diff
@@ -12,6 +12,179 @@ end
...
+end
+
+function BinningParameters(binstart,binend;numbins,kwargs...)
+    binwidth = (binend .- binstart) ./ numbins
+    return BinningParameters(binstart,binend,binwidth;kwargs...)
+end
+
+# Integrate over one or more axes of the histogram by setting the number of bins
+# in that axis to 1
+function integrate_params!(params::BinningParameters;axis)
```
ddahlbom, line 76: Perhaps `integrate_parameters!`, since "parameters" is written out in full elsewhere. Or maybe `integrate_axes!` is more precise. That it is operating on binning parameters is implied by the argument type.

src/StructureFactors/DataRetrieval.jl
```diff
@@ -12,6 +12,179 @@ end
...
+                inMiddle = true
+            end
+        end
+        @printf(io,"] (Δ = %.3f)", params.binwidth[k]/norm(params.covectors[k,:]))
+        println()
+    end
+end
+
+# Support numbins as a (virtual) property, even though only the binwidth is stored
+Base.getproperty(params::BinningParameters, sym::Symbol) = sym == :numbins ? round.(Int64,(params.binend .- params.binstart) ./ params.binwidth) : getfield(params,sym)
```
ddahlbom, line 53: This is a neat pattern. I hadn't seen it before.

src/StructureFactors/DataRetrieval.jl
```diff
@@ -12,6 +12,179 @@ end
...
+    for k in axis
+        nbins = [params.numbins.data...]
+        nbins[k] = 1
+        params.numbins = SVector{4}(nbins)
+    end
+end
+
+# This places one histogram bin around each possible Sunny scattering vector.
+# This is the finest possible binning without creating bins with zero scattering vectors in them.
+function unit_res_binning_parameters(sf::StructureFactor)
```
ddahlbom, line 86: I know these function names are getting long, but maybe write out `resolution` in full.

ddahlbom (commented):
This looks great. Thanks for the PR.

I do think `DataRetrieval.jl` is the right place for it. We may expand this into a folder with more fine-grained divisions in the future.

I was a bit skeptical on the call as to whether all the components should be stored in a single covector (as opposed to explicitly treating q and omega separately, as in the rest of the code), but it does result in cleaner code even if there are no relativistic transformations.

I guess you'll have to write a different `binned_histogram(sw:SpinWaveTheory, ...)`. Out of curiosity, what will be the approach there? Will you perform some kind of sampling/integration over the bins?

I put some (trivial) comments in the code.

src/StructureFactors/DataRetrieval.jl
Lazersmoke: Yep! One slight issue is that code like `params.numbins[4] = 1` doesn't work how you would expect (it's a no-op). I don't know if that's a deal breaker for this interface, but it's papered over by `integrate_params!` currently.

Lazersmoke:
Additional TODO on this, per discussion: need to incorporate broadening into the inner histogram loop, same as [LSWT code](https://github.com/SunnySuite/Sunny.jl/blob/6d5158613af37272f17ea647158c79716ab0f1eb/src/SpinWaveTheory/SWTCalculations.jl#L503).

Lazersmoke:
@ddahlbom Currently, the binned intensities functions (linear and spherical) do not handle the form factors, contractors, and possibly `kT` correctly, since they bypass the logic in `intensities`. What do you + @kbarros recommend to do about the interface for this? I may just finish implementing the binning internals and then turn it over to y'all for making the final user interface arrangements for these parameters (e.g. following the google doc)

ddahlbom:
@Lazersmoke Sure, that sounds great. Just let us know when when you're done making the final touches. I'd be happy to incorporate the rest of the intensities logic afterwards.

Lazersmoke:
OK, the final touches are done :)

I've added examples of binned version to the FeI2 and powder averaging tutorials. Some of the plotting is a bit rough around the edges.

There's a remaining question about the exact behavior of the "partial bins" at the far edge of the histogram, sometimes the behavior is possibly not quite correct, but the issue is that it's a bit under-specified what should be done in that case. Mantid has a bunch of options on like every function call that specify what to do with them, seems like. It shouldn't be a big issue for most use cases.

Remaining items for @ddahlbom and co:
- Documentation is in comment format right now, not sure on best practices for that
- The contractor/ffdata/kT interface needs to be bolted on the front of everything in an appropriate way
- The interface for the `integrated_kernel` could be improved, as per discussion in slack
- There is some code duplication between the powder-averaged and linear histograms; the logic is somewhat subtle surrounding the energy binning so one has to be careful, but they could possibly be combined by passing a `get_coordinates` function which generalizes both `covectors *` and `norm(q)`. Relatedly: The `covectors` are in Q space by default, but Mantid is in k space (absolute units). It may be advantageous to be in k space by default for that reason.

ddahlbom:
Thanks @Lazersmoke ! This is a great addition. I'll take a closer look at the code soon. If there are any changes that seem like they should be made before merging, I will comment here.

(I suppose we may want to merge sooner rather than later to keep things in sync. I can complete the remaining tasks after this is integrated.)

Issue #75, opened 2023-05-18
Open boundary conditions


kbarros:
We should make it easy to enable periodic boundaries. To achieve this, one can use [`to_inhomogeneous`](https://sunnysuite.github.io/Sunny.jl/dev/library/#Sunny.to_inhomogeneous-Union{Tuple{System{N}},%20Tuple{N}}%20where%20N) and then disable all the interactions that wrap the boundaries by hand. We can provide a convenience function with that behavior.

PR #85, opened 2023-06-20
New function, `remove_periodicity`

kbarros:
Also extend `symmetry_equivalent_bonds` to include offset information


kbarros:
Resolves https://github.com/SunnySuite/Sunny.jl/issues/75

PR #88, opened 2023-07-11
Swt dipole

Hao-Phys:
This PR adds support for the renormalized linear-spin wave theory for the `:dipole` mode. Features including
- Arbitrary single-ion anisotropy with proper renormalization.
- Scalar-biquadratic interaction.


src/SpinWaveTheory/SpinWaveTheory.jl
```diff
@@ -75,78 +77,79 @@ function generate_local_sun_gens(sys :: System)
...
+        Nₘ = length(sys.dipoles)
+        s̃_mat = Array{ComplexF64, 4}(undef, 2, 2, 3, Nₘ)
+        T̃_mat = zeros(ComplexF64, 0, 0, 0)
+        Q̃_mat = zeros(ComplexF64, 0, 0, 0, 0)
+        Nₘ = length(sys.dipoles)
+        R  = zeros(Float64, 3, 3)
+        for atom = 1:Nₘ
+            θ, ϕ = dipole_to_angles(sys.dipoles[1, 1, 1, atom])
+            # U_mat = exp(-1im*ϕ*s_mat_N[3]) * exp(-1im*θ*s_mat_N[2])
+            # SO(3) rotation that aligns the quantization axis. Important: since we will project out bosons that correspond to multipolar fluctuations,
```
kbarros, line 125: I am curious to know more about this. How will the convention for R become important?

Hao-Phys: Here we want to ensure that the operators' second row/column in the local reference frame corresponds to the boson with `Delta S = 1`. Then we can project out the bosons corresponding to multipolar fluctuations for the exchange interaction. Doing so allows us to use the same code when constructing the spin wave Hamiltonian for the exchange interactions.

src/SpinWaveTheory/SpinWaveTheory.jl
```diff
@@ -75,78 +77,79 @@ function generate_local_sun_gens(sys :: System)
...
     sys = reshape_geometry_aux(sys, (1,1,1), cellsize_mag)
 
-    s̃_mat, T̃_mat, Q̃_mat = generate_local_sun_gens(sys)
+    # Computes the Stevens operator in the local reference frame and the SO(3) rotation matrix from global to local frame
+    # (:dipole mode only)
+    c′_coef = Vector{StevensExpansion}()
+    R_mat   = Vector{Mat3}()
+    if sys.mode == :SUN
+        s̃_mat, T̃_mat, Q̃_mat = generate_local_sun_gens(sys)
+    elseif sys.mode == :dipole
```
ddahlbom, line 113: I wonder if you could move the contents of this `elseif` branch to an individual function, like you did for the rotated SU(N) operators? (I guess this function would also have to return the rotate Stevens coefficients.)

src/SpinWaveTheory/SpinWaveTheory.jl
ddahlbom: What specifically was the thing you had to change to get it to behave the way you want?

src/SpinWaveTheory/SpinWaveTheory.jl
```diff
@@ -75,78 +77,79 @@ function generate_local_sun_gens(sys :: System)
...
+        Nₘ, Ns = length(sys.dipoles), sys.Ns[1] # number of magnetic atoms and dimension of Hilbert 
+        s_mat_N = spin_matrices(Ns)
+        Nₘ = length(sys.dipoles)
+        s̃_mat = Array{ComplexF64, 4}(undef, 2, 2, 3, Nₘ)
```
ddahlbom, line 117: Small question of style, but I think you can probably eliminate the `_mat`. It will make the code read a bit more like your note.

src/SpinWaveTheory/SpinWaveTheory.jl
```diff
@@ -75,78 +77,79 @@ function generate_local_sun_gens(sys :: System)
...
-                s̃_mat[:, :, μ, atom] = Hermitian(U_mat' * s_mat_N[μ] * U_mat)
-            end
-            for ν = 1:5
-                Q̃_mat[:, :, ν, atom] = Hermitian(U_mat' * Q_mat_N[ν] * U_mat)
-            end
-            T̃_mat[:, :, atom] = Hermitian(U_mat' * sys.interactions_union[atom].aniso.matrep * U_mat)
+    for atom = 1:Nₘ
+        U_mat[:, 1] = sys.coherents[1, 1, 1, atom]
+        U_mat[:, 2:N] = nullspace(U_mat[:, 1]')
+        @assert isapprox(U_mat * U_mat', I) "rotation matrix from (global frame to local frame) not unitary"
```
ddahlbom, line 89: Is there any reason this would ever fail, assuming that there is nothing wrong with the original coherent state? (I am wondering if we can remove the assertion test.)

src/SpinWaveTheory/SpinWaveTheory.jl
```diff
@@ -75,78 +77,79 @@ function generate_local_sun_gens(sys :: System)
...
+            # Note that R * (0, 0, 1) = normalize(sys.dipoles[1,1,1,atom]))
+            R[:] = [-sin(ϕ) -cos(ϕ)*cos(θ) cos(ϕ)*sin(θ);
+                     cos(ϕ) -sin(ϕ)*cos(θ) sin(ϕ)*sin(θ);
+                     0.0     sin(θ)        cos(θ)]
+            (; c2, c4, c6) = sys.interactions_union[atom].aniso.stvexp
+            SR  = Mat3(R)
+            # In Cristian's note, S̃ = R S, so here we should pass SR'
+            push!(R_mat, SR')
+            for μ = 1:3
+                s̃_mat[:, :, μ, atom] = Hermitian(rotate_operator(s_mat_N[μ], SR))[1:2, 1:2]
```
ddahlbom, line 136: Why is the indexing at the end necessary? (Isn't the result 2x2?)

src/SpinWaveTheory/SWTCalculations.jl
```diff
@@ -15,20 +18,14 @@ function swt_hamiltonian!(swt::SpinWaveTheory, k̃ :: Vector{Float64}, Hmat::Mat
...
+    # Below two lines are for dipole mode only
+    S = (Ns-1) / 2 # spin-S
+    biquad_res_factor = 1 - 1/S + 1/(4S^2) # rescaling factor for biquadratic interaction
+
     Nf = sys.mode == :SUN ? Ns-1 : 1
     N  = Nf + 1
     L  = Nf * Nm
     @assert size(Hmat) == (2*L, 2*L)
-    # scaling factor (=1) if in the fundamental representation
-    M = sys.mode == :SUN ? 1 : (Ns-1)
```
ddahlbom, line 23: Sorry if I missed this. Why to the `M` scalings go away in your revision?

src/SpinWaveTheory/SWTCalculations.jl
```diff
@@ -3,7 +3,10 @@
...
 
 @inline δ(x, y) = ==(x, y) # my delta function
-
+# The "metric" of scalar biquad interaction. Here we are using the following identity:
+# (𝐒ᵢ⋅𝐒ⱼ)² = -(𝐒ᵢ⋅𝐒ⱼ)/2 + ∑ₐ (OᵢᵃOⱼᵃ)/2, a=4,…,8, 
+# where the definition of Oᵢᵃ is given in Appendix B of *Phys. Rev. B 104, 104409*
+const biquad_metric = 1/2 * diagm([-1, -1, -1, 1, 1, 1, 1, 1])
 
 """
     generate_ham_lswt!
```
ddahlbom, line 12: I wonder if this function could be broken down a little bit. For example, there could be functions`set_swt_pair_exchange!`, `set_swt_quadratic_exchange!`, `set_swt_sun_aniso`, `set_swt_dipole_aniso` or similar.

src/SpinWaveTheory/SpinWaveTheory.jl
```diff
@@ -17,7 +17,9 @@ struct SpinWaveTheory
     sys   :: System
     s̃_mat :: Array{ComplexF64, 4}  # dipole operators
```
ddahlbom, line 18: I suppose we may add a list of observables here (for constructing the relevant `Avec` in intensities). But that can come after merging. (Doesn't apply to dipole mode.)

src/SpinWaveTheory/SpinWaveTheory.jl
```diff
@@ -75,78 +77,79 @@ function generate_local_sun_gens(sys :: System)
...
+        Nₘ, Ns = length(sys.dipoles), sys.Ns[1] # number of magnetic atoms and dimension of Hilbert 
+        s_mat_N = spin_matrices(Ns)
+        Nₘ = length(sys.dipoles)
+        s̃_mat = Array{ComplexF64, 4}(undef, 2, 2, 3, Nₘ)
+        T̃_mat = zeros(ComplexF64, 0, 0, 0)
+        Q̃_mat = zeros(ComplexF64, 0, 0, 0, 0)
+        Nₘ = length(sys.dipoles)
+        R  = zeros(Float64, 3, 3)
+        for atom = 1:Nₘ
+            θ, ϕ = dipole_to_angles(sys.dipoles[1, 1, 1, atom])
```
ddahlbom, line 123: This function, `dipole_to_angles`, presumably because of singularities in the inverse trig functions. Does it ever fail? Can we put in conditions instead that handle the bad situations?

ddahlbom (commented):
Thanks Hao! It's exciting to have this capability. I just put in a few comments and questions.

src/SpinWaveTheory/SpinWaveTheory.jl
Hao-Phys: Hi David. This is a great suggestion. I will clean this part up.

src/SpinWaveTheory/SpinWaveTheory.jl
Hao-Phys: Good point!

src/SpinWaveTheory/SpinWaveTheory.jl
Hao-Phys: I think we can get rid of this assertion. Thanks!

src/SpinWaveTheory/SpinWaveTheory.jl
Hao-Phys: This is connected to the next question you are asking. Previously, I used the 2x2 matrix for the `:dipole` mode. Therefore, I must include the `M` scaling factor. Now I am using the (2S+1)x(2S+1) matrix and keeping only the first two columns/rows. In this way, the `M` factor is included. The revision looks nicer to me.

src/SpinWaveTheory/SWTCalculations.jl
Hao-Phys: See above

src/SpinWaveTheory/SWTCalculations.jl
```diff
@@ -415,8 +469,6 @@ function dssf(swt::SpinWaveTheory, qs)
     Nf = sys.mode == :SUN ? Ns-1 : 1
     N  = Nf + 1
     nmodes  = Nf * Nm 
-    M = sys.mode == :SUN ? 1 : (Ns-1) # scaling factor (=1) if in the fundamental representation
-    sqrt_M = √M
```
Lazersmoke, line 419: Side note: this change will need to be propagated to the relevant `intensity_formula` function (which cannibalizes `dssf`) during the merge

src/SpinWaveTheory/SpinWaveTheory.jl
kbarros: I asked Hao about this in person, and I think there are some subtleties. Maybe we should all have a conversation together.

src/SpinWaveTheory/SpinWaveTheory.jl
kbarros: I had this same concern. Upon closer investigation, it looks like the azimuthal angle `ϕ` is defined with `atan(x, y)`, and this promises to always return a number in $[-\pi, \pi]$. I.e., it makes some particular choice in cases where the azimuthal angle is ambiguous. Maybe we should still verify that the LSWT maintains the correct behavior in this ambiguous case (is it spins aligned with z?)

src/SpinWaveTheory/SWTCalculations.jl
Hao-Phys: Hi Sam. Yes, this is true. We should keep this in mind.

src/SpinWaveTheory/SpinWaveTheory.jl
Hao-Phys: For now, let's keep the `_mat`. We will have a more general way to define the generators later (for entangled units).

src/SpinWaveTheory/SpinWaveTheory.jl
ddahlbom: Gotcha, I see. Thanks for explaining.

Lazersmoke:
Closing this for now; if there are more updates (e.g. regarding `dipole_to_angles` or custom observables) they can be made on this branch and we'll re-open the PR, or made on a new branch which is up-to-date with `main`.

PR #90, opened 2023-07-13
Merge prototype version 0.5.0 staging branch into main

Lazersmoke:
This is to collect and resolve all the merge conflicts that are about to result from the various great changes everyone is working on :)

Changes incorporated so far are listed below:

---------

* Implemented `:dipole` mode for LSWT via @Hao-Phys  closes #88 

* Add `intensities_binned`

* Refactored interface for intensity calculations

* Add `load_nxs_binning_parameters`

* Refactor file structure

* Merge SWT dipole calcs with intensity_formula

---------


src/Sunny.jl
```diff
@@ -1,6 +1,7 @@
 module Sunny
 
 using LinearAlgebra
+using LinearMaps
```
kbarros, line 4: Please use `import LinearMaps` instead, and then qualify references, e.g. `LinearMaps.x`.

src/StructureFactors/StructureFactors.jl
```diff
@@ -53,35 +114,85 @@ function StructureFactor(sys::System{N}; Δt, nω, measperiod,
...
+        # projections (SU(N) mode) or components (dipole mode)
+        observable_ixs = Dict(:Sx => 1,:Sy => 2,:Sz => 3)
+        if N == 0
+            dipole_component(i) = FunctionMap{Float64}(s -> s[i],1,3)
+            observables = dipole_component.([1,2,3])
+        else
+            # SQTODO: Make this use the more optimized expected_spin function
+            # Doing this will also, by necessity, allow users to make the same
+            # type of optimization for their vector-valued observables.
+            observables = LinearMap{ComplexF64}.(spin_matrices(N))
```
kbarros, line 128: What is the motivation for using this instead of a dense `Hermitian{ComplexF64}`? If all else is equal, I'd prefer the explicit option. One could imagine that `LinearMap` could be useful for optimization if we have a very fast matrix-vector product available, but will that be the case in our intended applications?

Project.toml
```diff
@@ -12,8 +13,10 @@ FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 FilePathsBase = "48062228-2e41-5def-b9a4-89aafe57970f"
 Inflate = "d25df0c9-e2be-5dd7-82c8-3ad0b3e990b9"
 Interpolations = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LinearMaps = "7a12625a-238d-50fd-b39a-03d52299707e"
```
kbarros, line 19: Every additional package will affect load time. It seems like JLD2 will bring a lot of value, but maybe we can work around LinearMaps?

src/Sunny.jl
```diff
@@ -89,14 +92,30 @@ include("StructureFactors/StructureFactors.jl")
...
+    rlu_to_absolute_units!,
+    intensities_binned, slice_2D_binning_parameters, axes_bincenters,
+    connected_path_bins, count_bins
+
+
+include("Intensities/PowderAveraging.jl")
+export powder_averaged_bins, powder_average
+
+include("Intensities/ExperimentData.jl")
+export load_nxs, generate_mantid_script_from_binning_parameters
```
kbarros, line 118: This is a long list of exports. Have all of them been carefully considered? (I'm not familiar with these functions in particular, so please take this as a general comment).

Context: The export list is part of the package's SemVer guarantee. That is, there is an implicit promise not to change the behavior of functions if they are exported. If any of these functions still be evolving, we can hide them from the export list, while access is still possible through the syntax `Sunny.my_unexported_function`.

kbarros:
Great work Sam! I put some comments above for you to consider, but David's review will be more important.

src/StructureFactors/StructureFactors.jl
Lazersmoke: The main reason is that this (should) compile to the same thing as it did before this change specifically for dipole mode. A linear map `s -> s[i]` is better than multiplying by a row vector IMO. It's also nicer for advanced users to specify a linear map, especially in sun mode

kbarros: Do you have examples in mind of linear maps that are not explicit matrices?

kbarros: BTW, one place this might show up is in LSWT where we can in principle construct unitaries `exp(I theta S)` as a polynomial of sparse spin operators S. But maybe that type of thing never appears for observables.

Project.toml
kbarros: LinearMaps seems relatively lightweight: 
https://juliahub.com/ui/Packages/LinearMaps/bdfwy/3.10.2?page=1

If you think it will make the interface more convenient then I'm on OK with including this package.

examples/powder_averaging.jl
```diff
@@ -75,14 +75,13 @@ fig
...
 # just demonstrated above, and then averaged to produce a single vector of
 # energy-intensities for each `r`. Note that our `powder_average` function
-# passes most of its keywords through to [`intensities`](@ref), so it can be
-# given `kT`, `formfactors`, etc., and these parameters will be applied to the
-# calculation.
+# passes most of its keywords through to [`intensities_interpolated`](@ref), so it can be
+# given an [`intensity_formula`](@ref).
 
-function powder_average(sf, rs, density; η=0.1, mode=:perp, kwargs...)
+function powder_average(sf, rs, density; η=0.1, kwargs...)
```
ddahlbom, line 84: We should probably just use the built-in function if we're now exporting it. I can update this.

ddahlbom (commented):
Thanks for putting this all together. It looks great to me. I added a few minor comments in the code.

Project.toml
Lazersmoke: Yep, it's lightweight and in some sense the "correct" way to do certain kinds of linear algebra in Julia. We're bound to need it at some point

src/Sunny.jl
Lazersmoke: > Context: The export list is part of the package's SemVer guarantee. That is, there is an implicit promise not to change the behavior of functions if they are exported. If any of these functions may still be evolving, we can hide them from the export list, while access is still possible through the syntax `Sunny.my_unexported_function`.

I've just gone through and cleaned up the exports a bit. In terms of SemVer guarantee, we need to settle the units of the `BinningParameters` before unleashing this on the world, but once it is settled, `BinningParameters`, `count_bins`, `axes_bincenters`, `rlu_to_absolute_units!` (or its replacement)

Lazersmoke: @ddahlbom I think that's right re: `powder_average_interpolated`, since that function can (and is) implemented in terms of `intensities_interpolated`, whereas `powder_average_binned` is impossible to implement correctly in terms of `intensities_binned`. I vote un-export `powder_average_interpolated` but mention its existence in the Example

src/Intensities/Binning.jl
```diff
@@ -0,0 +1,438 @@
...
+- The bin width is `binwidth`
+- The last bin contains `binend`
+- There are no "partial bins;" the last bin may contain values greater than `binend`. C.f. [`count_bins`](@ref).
+
+A `value` can be binned by computing its bin index:
+
+    coords = covectors * value
+    bin_ix = 1 .+ floor.(Int64,(coords .- binstart) ./ binwidth)
+"""
+mutable struct BinningParameters
```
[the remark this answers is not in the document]
Lazersmoke: This is a very important point. I think what actually needs to happen is to make the `BinningParmaeters` in terms of physical (absolute) units instead of R.L.U., which are problem- and lattice-dependent (i.e. they aren't a real unit in any meaningful sense). As a bonus, no conversion is needed to go to Mantid.

src/Intensities/Interpolation.jl
```diff
@@ -0,0 +1,296 @@
...
+
+- `interpolation`: Since ``𝒮(𝐪, ω)`` is calculated on a finite lattice, data
+    is only available at discrete wave vectors. By default, Sunny will round a
+    requested `q` to the nearest available wave vector. Linear interpolation can
+    be applied by setting `interpolation=:linear`.
+- `negative_energies`: If set to `true`, Sunny will return the periodic
+    extension of the energy axis. Most users will not want this.
+"""
+function intensities_interpolated(sf::StructureFactor, qs;
+    formula = intensity_formula(sf,:perp) :: ClassicalIntensityFormula,
```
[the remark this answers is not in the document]
Lazersmoke: It's correct that the `IntensityFormula` types are never exported, and that they are opaque to the user. I've been defining the types right above the `intensity_formula` used to construct them, e.g. this one is in `src/StructureFactors/DataRetrieval.jl`.

src/Intensities/ElementContraction.jl
```diff
@@ -0,0 +1,183 @@
...
+required_correlations(dipoleinfo::DipoleFactor) = dipoleinfo.indices
+required_correlations(eleminfo::Element) = [eleminfo.index]
+required_correlations(::FullTensor{NCorr}) where NCorr = 1:NCorr
+
+
+################################################################################
+# Contraction utils
+################################################################################
+Base.zeros(::Contraction{T}, dims...) where T = zeros(T, dims...)
+contraction_return_type(::Contraction{T}) where T = T
```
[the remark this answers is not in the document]
Lazersmoke: Yep, that way is better

src/StructureFactors/StructureFactors.jl
```diff
@@ -21,11 +21,51 @@ struct StructureFactor{N}
     processtraj! :: Function              # Function to perform post-processing on sample trajectories
 end
 
-function Base.show(io::IO, ::MIME"text/plain", sf::StructureFactor)
-    modename = sf.dipole_corrs ? "Dipole correlations" : "Custom correlations"
-    printstyled(io, "StructureFactor [$modename]\n"; bold=true, color=:underline)
+function Base.show(io::IO, sf::StructureFactor{N}) where N
```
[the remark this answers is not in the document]
Lazersmoke: I like this idea for the reason that I was initially confused by the fact that no part of using spin wave theory involves constructing a `StructureFactor`

src/Intensities/PowderAveraging.jl
```diff
@@ -0,0 +1,129 @@
...
+
+Similar to [`powder_average`](@ref), but the data is binned instead of interpolated.
+Also similar to [`intensities_binned`](@ref), but the histogram x-axis is `|k|` in absolute units, which is a nonlinear function of `kx`,`ky`,`kz`. The y-axis is energy.
+
+Binning parameters are specified as tuples `(start,end,bin_width)`,
+e.g. `radial_binning_parameters = (0,6π,6π/55)`.
+
+Energy broadening is supported in the same way as `intensities_binned`.
+"""
+function powder_averaged_bins(sf::StructureFactor, radial_binning_parameters;
```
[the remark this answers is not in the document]
Lazersmoke: Yep, this is good

src/Sunny.jl
```diff
@@ -89,14 +92,30 @@ include("StructureFactors/StructureFactors.jl")
...
+
+include("Intensities/Interpolation.jl")
+export intensities_interpolated, instant_intensities_interpolated, connected_path
+
+include("Intensities/Binning.jl")
+export intensities_binned, instant_intensities_interpolated, connected_path_bins,
+    BinningParameters, integrate_axes!, unit_resolution_binning_parameters,
+    rlu_to_absolute_units!,
+    intensities_binned, slice_2D_binning_parameters, axes_bincenters,
+    connected_path_bins, count_bins
```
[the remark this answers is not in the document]
Lazersmoke: and `intensities_binned` lol. I've cleaned it up now

src/SpinWaveTheory/kpm.jl
```diff
@@ -0,0 +1,45 @@
...
+        precursor = repeat(' ', textwidth(intensity_equals))
+        println(io,precursor,formula_lines[i])
+    end
+    println(io,"P = $(formula.P), kT = $(formula.kT), σ = $(formula.σ)")
+end
+
+
+function intensity_formula_kpm(f::Function,swt::SpinWaveTheory,corr_ix::AbstractVector{Int64}; P =, return_type = Float64, string_formula = "f(Q,ω,S{α,β}[ix_q,ix_ω])")
+
+    error("KPM not yet implemented")
```
Lazersmoke, line 34: The plan is this gets clobbered in the merge with kpm branch eventually

src/Plotting.jl
```diff
@@ -253,6 +253,29 @@ function plot_spins(sys::System; linecolor=:grey, arrowcolor=:red,
...
+        vecs = Makie.Vec3f0.(sys_origin.dipoles[:])
+        Makie.arrows!(
+            ax, pts, vecs;
+            linewidth, arrowsize,
+            lengthscale=arrowlength,
+            # Original system is transparent
+            arrowcolor=(arrowcolor,0.03), linecolor = (linecolor,0.03),
+            kwargs...
+        )
+    end
```
Lazersmoke, line 277: This may have been better as a separate PR, but it does this:

image

which I think is neat and resolves the confusion I had looking at this example for the longest time (I thought "supercell" meant that each of the four spins was meant to represent a cluster of spins in the original system, rather than a sublattice)

src/Sunny.jl
ddahlbom: I'm fine with not exporting it. (Reference to it should then be eliminated from the doc string for the binned version.)

src/Intensities/Binning.jl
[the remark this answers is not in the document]
ddahlbom: Sounds good. Doing all the binning code consistently in terms of absolute units makes sense to me.

src/Intensities/Interpolation.jl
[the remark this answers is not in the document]
ddahlbom: Oh, yeah. Thanks. I overlooked that.

src/StructureFactors/StructureFactors.jl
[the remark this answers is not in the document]
Lazersmoke: Yep ok agreed. I think this is a good idea in general, but out of scope for the PR :)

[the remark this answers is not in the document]
Lazersmoke: Yep absolutely! I'm nearly done converting everything to use absolute units everywhere, it's quite thorny to track down every place where RLU has been assumed.

Lazersmoke:
Assuming the CI goes ok, I'll rebase/squash & merge this later today, hopefully in a sensible way

src/StructureFactors/StructureFactors.jl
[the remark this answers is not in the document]
ddahlbom: Sounds good. Just to be clear, when you say "everywhere," do you mean in the binning functions? I wouldn't change the default behavior of the SWT or classical interfaces. RLUs make a lot of sense as the default units elsewhere, I think. (For example, it's much more meaningful to say that the ordering wave vector is (pi, pi, pi) than to give the same information in absolute units.)

PR #92, opened 2023-07-13
Add KPM to main

Lazersmoke:
This feature is under active development by Harry, so I assume it's not ready yet. This PR is so that it is not forgotten!


Lazersmoke:
Also: per a discussion with Martin today, a third kind of intensity formula, `KPMIntensityFormula` will need to be added, since KPM responds to `S(q,w)` queries instead of `(disp(q),intensity(q))` queries. I can create this from the `KPMintensities` and `KPMdssf` functions if/when they are stabilized.

kbarros:
> Also: per a discussion with Martin today, a third kind of intensity formula, `KPMIntensityFormula` will need to be added, since KPM responds to `S(q,w)` queries instead of `(disp(q),intensity(q))` queries.

I am not sure about the details, but this sounds familiar to the situation from classical dynamics -- in both cases, we will be querying `S(q,w)` for a list of `q` values. Could the UI be shared?

Lazersmoke:
> in both cases, we will be querying `S(q,w)` for a list of `q` values. Could the UI be shared?

The UX *is* shared (in the sense that its just `intensity_formula` in all cases), but they are doing fundamentally different things.

Classical: list of indexes describing `q`s from the list of available `q`s. If a `q` is outside that list, then interpolation needs to happen, `intensities_interpolation`

KPM: list of any, arbitrary `q`s, with no interpolation. Then, we wouldn't want users to call `intensities_interpolated` because that's misleading. So it should probably be called either plain `intensities` or `intensities_kpm` if we don't want people confused why `intensities(::StructureFactor)` isn't defined. (c.f. LSWT uses `intensities_broadened` and `intensities_bands` depending if there is broadening)

What can *maybe* be done is to define `intensities_binned(::KPM)` (but only if we can analytically integrate the KPM `Intensity(q,w)` function over small finite volumes `(q+dq,w+dw)`, @hlane33 ?), and actually this would be highly desirable for inverse problem if its fast.

Lazersmoke:
FYI on this: Probably the Cairo dependency at least should be removed before merging. That's just for the examples, which can go in SunnyTutorials instead presumably? Or ported to GLMakie and left here

kbarros:
Before we merge this I think @hlane33 should reapply the commits so that they are correctly credited as the author (coauthor ddahlbom).

kbarros:
Development for this feature has moved to #289.

PR #93, opened 2023-07-13
Refactor single ion anisotropies

kbarros:
- Replace `set_anisotropy!` with `set_onsite!`. The new function expects a matrix built from `stevens_operators` or  `spin_operators`. Provide `print_stevens_expansion` to decompose polynomials of spin operators into a linear combination of Stevens operators.

- Make DynamicPolynomials an optional dependency. This provides symbolic functions `print_classical_stevens_expansion` and `print_classical_spin_polynomial`.

- Functions to create tensor products of operators.


kbarros:
There are various breaking changes in this commit, but the main one is the replacement `set_anisotropy!` with `set_onsite!` (name TBD). The new function takes an explicit $N \times N$ matrix which can be built using, e.g. `spin_operators` or `stevens_operators`. DynamicPolynomials.jl becomes an optional dependency, and can provide certain symbolic analysis relevant to the large $S$ classical limit.

test/test_symmetry.jl
```diff
@@ -154,21 +154,22 @@ end
...
-        Λ = 𝒪[6,0]-(35/√8)*𝒪[6,3]+(77/8)*𝒪[6,6]
+        Λ = O[6,0]-(35/√8)*O[6,3]+(77/8)*O[6,6]
         Λ′ = rotate_operator(Λ, R)
         @test Sunny.is_anisotropy_valid(cryst, i, Λ′)
 
         latvecs = lattice_vectors(1.0, 1.1, 1.0, 90, 90, 90)
         cryst = Crystal(latvecs, [[0., 0., 0.]])
         # print_site(cryst, i)
-        Λ = randn()*(𝒪[6,0]-21𝒪[6,4]) + randn()*(𝒪[6,2]+(16/5)*𝒪[6,4]+(11/5)*𝒪[6,6])
+        Λ = randn()*(O[6,0]-21O[6,4]) + randn()*(O[6,2]+(16/5)*O[6,4]+(11/5)*O[6,6])
```
kbarros, line 172: Using `O` here (instead of `𝒪`) doesn't look great when juxtaposed with coefficients. The tradeoff is that some people's terminals won't render `𝒪` properly, so I was thinking it might be better to use plain ASCII in the main examples.

examples/powder_averaging.jl
```diff
@@ -7,7 +7,6 @@
 
 using Sunny, GLMakie
 using Statistics: mean
-#md Makie.inline!(true); #hide
```
Lazersmoke, line 10: Was this a no-op?

kbarros: It was previously needed as a workaround for this bug https://github.com/MakieOrg/Makie.jl/issues/2805.

test/test_symmetry.jl
Lazersmoke: Probably, users will make themselves clear by inserting `*` for clarity. Also consider that this form of juxtaposed coefficients is a special feature of julia, so people are already wary. This term: `(O[6,2]+(16/5)*O[6,4]+(11/5)*O[6,6])` looks fine to me, and even more fine with spaces `(O[6,2] + (16/5)*O[6,4] + (11/5)*O[6,6])`.

Definitely Unicode-optional is the way to go.

src/Operators/TensorOperators.jl
```diff
@@ -0,0 +1,126 @@
+# Produces a matrix representation of a tensor product of operators, C = A⊗B.
+# Like built-in `kron` but with permutation. Returns C_{acbd} = A_{ab} B_{cd}.
+function kron_operator(A::AbstractMatrix, B::AbstractMatrix)
```
Lazersmoke, line 3: Is this one of the kron-like matrix products defined by `LinearMaps.jl`? [link](https://julialinearalgebra.github.io/LinearMaps.jl/stable/types/#Base.kron-Tuple{LinearMap,%20LinearMap}). It may be worth investigating whether `LinearMaps` helps in anyway with these onsite operators, [c.f.](https://github.com/SunnySuite/Sunny.jl/blob/148a051b39dca3e880a39b9ea3d76fe12bfea001/src/StructureFactors/StructureFactors.jl#L122). Seems unlikely because they are always dense matrices.

Also, is it faster than `Base.permutedims` + `Base.kron`?

Lazersmoke:
Looks great! Good to merge as-is

ddahlbom (approved):
This all looks great to me too.

src/Operators/TensorOperators.jl
kbarros: Good catch. The kron from LinearMaps is semantically the same. It might be lazy, which would save work at construction time, but I suspect it wouldn't significantly reduce the total amount of work required (e.g. there will eventually be an SVD operation).

Eventually we should have a chat about the cost of constructing Hamiltonians in the context of inverse modeling. If optimization and LSWT becomes really fast, it could be that building the Hamiltonian is the dominant cost (e.g. symmetry analysis can sometimes be really slow). I have some thoughts to run by you.

PR #94, opened 2023-07-16
PairCoupling refactors

kbarros:
Internal cleanups to unify treatment of pair couplings. In particular, each atom now stores only a single list of PairCouplings, which may contain bilinear (scalar or matrix) and biquadratic (scalar-only) exchange.

The function `set_biquadratic!` has been removed. Instead, call `set_exchange!` with the keyword argument `biquad`.

The changes to the internal datastructures will require updates to LSWT, which should also be incorporated into the branches of @Hao-Phys and @hlane33.

### DIFF
--- a/src/SpinWaveTheory/SWTCalculations.jl
+++ b/src/SpinWaveTheory/SWTCalculations.jl
@@ -26,6 +26,16 @@ function swt_hamiltonian!(swt::SpinWaveTheory, k̃ :: Vector{Float64}, Hmat::Mat
     N  = Nf + 1
     L  = Nf * Nm
     @assert size(Hmat) == (2*L, 2*L)
+    # scaling factor (=1) if in the fundamental representation
+    M = sys.mode == :SUN ? 1 : (Ns-1)
+    no_single_ion = isempty(swt.sys.interactions_union[1].onsite.matrep)
+
+    # the "metric" of scalar biquad interaction. Here we are using the following identity:
+    # (𝐒ᵢ⋅𝐒ⱼ)² = -(𝐒ᵢ⋅𝐒ⱼ)/2 + ∑ₐ (OᵢᵃOⱼᵃ)/2, a=4,…,8
+    # where the definition of Oᵢᵃ is given in Appendix B of *Phys. Rev. B 104, 104409*
+    # Note: this is only valid for the `:dipole` mode, for `:SUN` mode, we consider 
+    # different implementations
+    biquad_metric = 1/2 * diagm([-1, -1, -1, 1/M, 1/M, 1/M, 1/M, 1/M])
 
     for k̃ᵢ in k̃
         (k̃ᵢ < 0.0 || k̃ᵢ ≥ 1.0) && throw("k̃ outside [0, 1) range")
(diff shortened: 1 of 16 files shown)


src/SpinWaveTheory/SWTCalculations.jl
```diff
@@ -59,60 +59,18 @@ function swt_hamiltonian!(swt::SpinWaveTheory, k̃ :: Vector{Float64}, Hmat::Mat
...
-                    Hmat12[sub_i_M1*Nf+mM1, sub_j_M1*Nf+nM1] += 0.5 * c5 * phase
-                    Hmat12[sub_j_M1*Nf+nM1, sub_i_M1*Nf+mM1] += 0.5 * c5 * cphase
-                    Hmat21[sub_i_M1*Nf+mM1, sub_j_M1*Nf+nM1] += 0.5 * c6 * phase
-                    Hmat21[sub_j_M1*Nf+nM1, sub_i_M1*Nf+mM1] += 0.5 * c6 * cphase
-                end
+            ### Bilinear exchange
+            if coupling.bilin isa Number
+                J = Mat3(coupling.bilin*I)
+            else
+                J = coupling.bilin
```
Lazersmoke, line 71: It seems like a problem to me that these lines are just removed now? Is there supposed to be further changes re-introducing the behavior?

src/System/Interactions.jl
```diff
@@ -227,45 +223,42 @@ function energy_aux(sys::System{N}, ints::Interactions, i::Int, cells, foreachbo
...
-        E += J * dot(sᵢ, sⱼ)
-    end
 
-    # Quadratic exchange matrix
-    foreachbond(ints.exchange) do J, site1, site2
-        sᵢ = dipoles[site1]
-        sⱼ = dipoles[site2]
-        E += dot(sᵢ, J, sⱼ)
-    end
+        if coupling.bilin isa Float64
```
Lazersmoke, line 240: Dynamic dispatch? Benchmark. Also if you run this in `@profview` and see red in the top row on this line, that's an easy way to check if this is a meaningful problem

Lazersmoke: Actually scratch that, this branch can be removed entirely! `dot(v,::Float64,v)` is defined, let `dot` branch internally! :) @kbarros

src/System/Interactions.jl
```diff
@@ -312,73 +305,70 @@ function set_forces_aux!(B, dipoles::Array{Vec3, 4}, ints::Interactions, sys::Sy
...
-        sⱼ = dipoles[site2]
+        if coupling.bilin isa Float64
+            J = coupling.bilin::Float64
+            B[site1] -= J  * sⱼ
+            B[site2] -= J' * sᵢ
+        else
+            J = coupling.bilin::Mat3
+            B[site1] -= J  * sⱼ
+            B[site2] -= J' * sᵢ
+        end
```
Lazersmoke, line 324: Ditto, these code are the same thing and is type stable, don't branch explicitly

Lazersmoke: OH, I see you're restricting the possible options it has to deal with. I think it's probably still not needed...

src/System/PairExchange.jl
```diff
@@ -14,68 +14,27 @@ function bond_parity(bond)
...
-            # Add to list
-            coupling = Coupling(isculled, bond′, J)
-            push!(ints[i].biquad, coupling)
-        end
-
-        # Sort interactions so that non-culled bonds appear first
-        sort!(ints[i].biquad, by=c->c.isculled)
+# Convert J to Union{Float64, Mat3}
+function to_float_or_mat3(J)
+    if J isa Number || isapprox(J, J[1] * I, atol=1e-12)
```
Lazersmoke, line 19: Why is this not a floating point equality check? Is there some situation where we want to round a user's matrix?

src/System/PairExchange.jl
```diff
@@ -133,36 +93,15 @@ function set_exchange!(sys::System{N}, J, bond::Bond) where N
     end
 
     # Print a warning if an interaction already exists for bond
-    if any(x -> x.bond == bond, vcat(ints[bond.i].heisen, ints[bond.i].exchange))
+    if any(x -> x.bond == bond, ints[bond.i].pair)
```
Lazersmoke, line 96: Feature request; flag to disable this warning. I use this to re-run the forwards problem with different `J` each time and currently no way to turn this off easily

src/System/Interactions.jl
Lazersmoke: Another possibility: This is what `UniformScaling` was designed for, right? there's also `dot(v,::UniformScaling,v)`

Lazersmoke:
Looks good! I think the type stuff (`Float64` vs `UniformScaling`) can even be addressed at a later time, the interface probably shouldn't change based on that

src/SpinWaveTheory/SWTCalculations.jl
kbarros: The behavior should be the same (if J is stored as a scalar, it has the meaning of a scalar times the 3x3 identity matrix). The priority here is code de-duplication. Performance optimization will be a separate pass.

src/System/Interactions.jl
kbarros: In the Julia docs there was warning that a Union{Float64, Mat3} cannot be stored on the stack, and would get heap allocated instead. That is the reason to never load coupling.bilin into a local variable, but instead put it behind the `isa` conditionals. Those statements are treated specially by the optimizer at inference time.

src/System/Interactions.jl
kbarros: This code was the result of some experiments and reading the docs. Maybe easiest to discuss on slack.

src/System/PairExchange.jl
```diff
@@ -84,11 +43,16 @@ interactions on these bonds will be overwritten. The parameter `bond` has the
...
+``𝐃⋅(𝐒_i×𝐒_j)``.
 
-As a convenience, `dmvec(D)` can be used to construct the antisymmetric part of
-the exchange, where `D` is the Dzyaloshinskii-Moriya pseudo-vector. The
-resulting interaction will be ``𝐃⋅(𝐒_i×𝐒_j)``.
+The optional parameter `biquad` defines the strength ``b`` for scalar
+biquadratic interactions of the form ``b (𝐒_i⋅𝐒_j)²`` For systems restricted
+to dipoles, ``b`` will be automatically renormalized for maximum consistency
+with the more variationally accurate SU(_N_) mode. This renormalization
+introduces also a correction to the quadratic part of the exchange.
```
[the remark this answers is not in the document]
kbarros: What you're suggesting is a general biquadratic interaction, which we don't support yet. There are currently no plans to support it in dipole mode because it would be fairly complicated (a generalization of the renormalization scheme here would need to be derived, https://arxiv.org/abs/2304.03874). The plan is to generalize this entire thing for SU(N) mode so people can build arbitrary spin polynomials, not restricted to biquadratic.

src/System/PairExchange.jl
Lazersmoke: No, I mean why `isapprox` instead of `==`. I see now that this behavior was already here before the PR, but why?

src/System/Interactions.jl
kbarros: After some experimentation, the changes that Sam suggested seem to work well without heap allocation. Specifically, in this PR, it is possible to do:
```
J = coupling.bilin
dot(v, J, v)
```
and the compiler can avoid heap allocation.

src/System/PairExchange.jl
kbarros: Let's avoid floating point equality checks wherever possible.

As a concrete example, suppose we are generating `J` by a rotation: `J = R * J0 * R'`, where `J0` happens to be exactly diagonal. Mathematically, we should have `R * R' = I`, but numerically this might only be approximate. The approximate equality check will correctly infer that `J` is _intended_ to be diagonal.

src/System/PairExchange.jl
kbarros: Good idea. Per Slack discussion, let's address this in a follow-up PR using https://docs.julialang.org/en/v1/stdlib/Logging/

src/System/Interactions.jl
kbarros: I have updated the PR to use multiplication of `Union{Float64,Mat3}`, which Julia seems to optimize well. Also, if there were type instability, the JET.jl unit test seem to be good about finding it.

src/System/PairExchange.jl
kbarros: Also, from the users perspective, the stakes are:

- change in numerical result at the level of 1e-12, presumably not a big deal in most contexts.
- possibly significant speedup

kbarros: After thinking about it more, maybe we should log a warning if this situation is detected.

kbarros: Hmm, I think I agree with you now 😆. The situation is rare, performance difference will likely be small, and using a full Mat3 would never be wrong.

Lazersmoke: I actually quite like the warning idea! Behavior is to use whatever they give us, but if it's numerically close to diagonal, a warning/suggestion that they could make it exactly diagonal for better performance may be warranted!

PR #98, opened 2023-07-19
Add VTK Export

Lazersmoke:
This enables users to export histograms and `System`s (as dipoles + S(Q) intensity histograms) to a ParaView-compatible format.

This is about as far as I'm planning on pushing the integration here, since ParaView is a bit clunky for our use case, and absolutely does not handle 4D whatsoever. Mainly the volumetric viewing (& slicing, not demo'd here) are quite nice.


kbarros:
Thanks, this looks well put together.

Is the main purpose of this to do volumetric rendering? If so, I wonder whether GLMakie could serve a similar purpose? It provides a few different rendering algorithms: https://docs.makie.org/stable/examples/plotting_functions/volume/
An advantage of having it ourselves in Sunny is that we can then draw the 1st BZ using Brillouin.jl, and add more features later.

Another consideration is space requirements for the images used in the docs. Most of them are ~20k, which is great. The Paraview render is ~300k, which starts to get large enough maybe we should think carefully. I bring this up because (I think) once this gets committed, the images will always need be downloaded on `git clone`, even if we later remove the feature. (Is that correct?)

Lazersmoke:
I've switched it out for a smaller `.jpg`. If we squash when we merge, the larger file will not be in the commit history of the main repo.

Re: BZ plotting and more, it's actually pretty straightforward to add extra structures/data like that to the exported file (few LOC each feature, supposing `Brillouin.jl` can export a list of vertices for example). And once it's in the ParaView universe, all possible features of ParaView [there are many] can be unleashed without us writing any code (i.e. it's useful for people who already have their tooling in place to get their hands on Sunny data in a standard format)

Certainly, GLMakie can still be used, and is closer to interactive, it's just less established/featureful

kbarros:
Looks good to me. Merge whenever you think it's ready.

PR #101, opened 2023-07-21
LSWT docs

kbarros:
Two new docs from Hao and Cristian. This will require some discussion and work before merging. This PR can serve as a forum for discussion and planning.

1. There is already stevens.lyx which overlaps somewhat with LSWT_dipole.lyx. Material from these two documents should be merged, and redundancies removed.
2. SunnySWT.lyx looks like it might still be evolving, and has extra materials besides just LSWT. We might want to extract a simplified subset of the material that focuses on SU(N) LSWT.
3. The two notes LSWT_dipole.lyx and SunnySWT.lyx were auto-converted from Latex to Lyx. This leads to some artifacts that should be cleaned up.

I am also wondering if we eventually want to convert these to Markdown format and include them in our online docs.

**Update**. Here is the [compiled PDF](https://github.com/user-attachments/files/24945828/hao-lswt-note.pdf) of Hao's LSWT note.

### DIFF
--- a/docs/lyx_notes/LSWT_dipole.lyx
+++ b/docs/lyx_notes/LSWT_dipole.lyx
@@ -1,0 +1,5360 @@
+#LyX 2.3 created this file. For more info see http://www.lyx.org/
+\lyxformat 544
+\begin_document
+\begin_header
+\save_transient_properties true
+\origin unavailable
+\textclass revtex4-2
+\begin_preamble
+
+
+\usepackage{mathrsfs}
+\usepackage{babel}
+\usepackage{color}
+\usepackage{graphicx}
+\newcommand{\inlinecode}{\texttt}
+\end_preamble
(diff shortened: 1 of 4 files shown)


Lazersmoke:
For the online docs: It might be appropriate to make a "Calculation Details" header (and put it after "Examples") to collect all the information people need to understand the calculations which are performed by Sunny:

- These LSWT notes (I envision the `src/SpinWaveTheory/SpinWaveTheory.jl` and/or `SWTCalculations.jl` [or more likely combining these into one file] as becoming Literate files used as source for the documentation page, with the notes adapted into markdown comments)
- Histogram tutorial
- What is currently the "Structure Factor Calculations" docs page
- A page about multipolar moments, e.g. using the spin-1 chain example from Bai et al
- etc

Issue #71, opened 2023-04-11
Procedure to quickly find local energy minimum


kbarros:
LSWT requires as input a spin configuration that minimizes the classical energy. Our tutorials currently suggest finding a local minimum by running a long $T=0$ dynamics. Instead, we should provide an `minimize_energy(sys)`  function that converges exponentially fast, using e.g. L-BFGS: http://julianlsolvers.github.io/Optim.jl/v0.9.3/algo/lbfgs/

For this we will need the negative energy gradient. In dipole mode, this is available as `forces(sys)`. In SU(N) mode, I think the negative gradient is given by $\mathfrak{H}_i \mathbf{Z}_i$. See Integrators.jl for the procedure to numerically calculate this gradient (the variable name is `HZ`).

PR #102, opened 2023-07-25
Complete removal of "force" and optimizer fixes

kbarros:
This should handle the cases of non-uniform dipole or coherent magnitudes.


ddahlbom (approved):
Just recording for posterity that I think this all looks great.

kbarros:
Fixes https://github.com/SunnySuite/Sunny.jl/issues/71.

PR #103, opened 2023-07-25
Use Logging for info and warning printouts

kbarros:
**UPDATED**

Only the first 5 site or pair coupling override warnings will be printed to the terminal. After that, the warnings will be suppressed.

Convert some `println` statements to `Logging.@info` and `@warn` statements. This is appropriate for providing a warning that the user wasn't expecting. Some functions, of the form `print_*` explicitly promise to print something, and here the `println()` statements seem appropriate.


Lazersmoke:
Something strange with git happening here, be careful when you merge!

But these changes look great. I think the default behavior should be to warn the user if they try to override an interaction, especially with the new way biquad is handled.

Lazersmoke:
There's a few more in https://github.com/search?q=repo%3ASunnySuite%2FSunny.jl%20println&type=code

but this is probably an update-as-needed situation

PR #104, opened 2023-07-25
Sampled correlations

ddahlbom:
This just changes the name `StructureFactor` to `SampledCorrelations`. It also eliminates the automatic generation of an initial sample when creating a `SampledCorrelations`, as was discussed some time back.

I am getting some new warnings when building the docs, but they seem to look fine when I inspect them.

There's not much here, but many lines are touched. Let me know if you see any problems.


docs/src/structure-factor.md
```diff
@@ -81,28 +75,27 @@ available after the calculation has completed.
...
+[`LocalSampler`](@ref). The statistical quality of the $𝒮^{αβ}(𝐪,ω)$ can be
+improved by repeatedly generating decorrelated spin configurations in `sys` and
+calling `add_sample!` on each configuration.
 
 The outline of typical use case might look like this:
 ```
-# Make a `StructureFactor` and calculate an initial sample
-sf = DynamicStructureFactor(sys; Δt=0.05, ωmax=10.0, nω=100) 
+# Make a `SampledCorrelations` and calculate an initial sample
+sc = SampledCorrelations(sys; Δt=0.05, ωmax=10.0, nω=100)
```
Lazersmoke, line 88: This no longer has an initial sample as suggested by the comment

docs/src/structure-factor.md
```diff
@@ -122,7 +115,7 @@ configurations. Information about calculating instantaneous data from a dynamic
 structure factor can be found in the following section.
 
 The basic usage for the instantaneous case is very similar to the dynamic case,
-except one calls `InstantStructureFactor` instead of `DynamicStructureFactor`.
+except one calls `InstantStructureFactor` instead of `SampledCorrelations`.
```
Lazersmoke, line 118: Unsure what to do about `InstantStructureFactor`

Lazersmoke: Seems weird to me that calling `InstantStructureFactor` builds you a `SampledCorrelations` (none of those words are the same anymore)

Lazersmoke:
Yep, the warnings are not your fault. It's some Documenter issue, has to do with the recently tweaked docs pipeline. I haven't found a way to resolve them, but I think they occur when you link from one docstring to another docstring.

docs/src/structure-factor.md
ddahlbom: Good catch. Thanks.

docs/src/structure-factor.md
ddahlbom: Agreed. I thought this was just where we left it in the thread yesterday.

I guess would could have kept the whole `Instant` vs `Dynamic` dichotomy: `SampledDynamicCorrelations` and `SampledInstantCorrelations`.

ddahlbom:
Gotcha. I think one warning will go away if a docstring is added to `export_vtk`.

docs/src/structure-factor.md
kbarros: How about `SampledCorrelations` and `InstantCorrelations`, where "sampling" refers to the dynamical sampling process?

kbarros: What are the primary use cases of `InstantCorrelations` right now? Also note the overlap with `print_wrapped_intensities`.

ddahlbom: Ok

kbarros:
Per @ddahlbom's comment:

> The motivation [for not automatically calling `add_sample!`] was that sometimes people want to do some post-processing of their trajectories before calculating correlations, so the first calculation would just be thrown out and therefore be wasteful. Xiaojian asked for this.

docs/src/versions.md
```diff
@@ -4,11 +4,20 @@ This version includes many **breaking changes**.
...
 
-Split `intensities` into calculation ([`intensity_formula`](@ref)) and presentation ([`intensities_interpolated`](@ref), [`intensities_binned`](@ref)). This is a **breaking change**, see the docs to migrate your code.
+Split `intensities` into calculation ([`intensity_formula`](@ref)) and
+presentation ([`intensities_interpolated`](@ref), [`intensities_binned`](@ref)).
+This is a **breaking change**, see the docs to migrate your code.
 
-Broadened support for custom observables in `StructureFactor` for use in `intensity_formula`.
+`StructureFactor` type renamed `SampledCorrelations`. An appropriate
+`SampledCorrelations` is created by calling either `dynamical_correlations` or
+`instant_correlations` instead of `DynamicStructureFactor` or
```
kbarros, line 13: Add hyperlinks,
```
[`SampledCorrelations`](@ref)
```
and similarly for `dynamical_correlations` and `instant_correlations`?

kbarros:
Looks good to me. Thanks David!

docs/src/versions.md
ddahlbom: Thanks. Added. (Note that the documentation to `SampledCorrelations` is basically just a stub and points to the docstrings for `dynamical_intensities` and `instant_intensities`. But I think it is useful to export `SampledCorrelations`, just because so many locations in the docs really want to refer to `SampledCorrelations` itself.)

docs/src/versions.md
```diff
@@ -46,7 +55,7 @@ can be resolved by passing an explicit `offset`.
 The function [`remove_periodicity!`](@ref) disables periodicity along specified
 dimensions.
 
-Rename `StaticStructureFactor` to [`InstantStructureFactor`](@ref).
+Rename `StaticStructureFactor` to [`InstantStructureFactor`].
```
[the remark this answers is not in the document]
ddahlbom: Yes. That line is describing an earlier version, I thought. Just keeping the record while removing the broken hyperlink.

PR #105, opened 2023-07-28
Implement several renamings

kbarros:
As described in versions.md:

```
Rename `polarize_spin!` to [`set_dipole!`](@ref) for consistency with
[`set_coherent!`](@ref). The behavior of the former function is unchanged: the
spin at a given site will still be polarized along the provided direction.

Rename `reshape_geometry` to [`reshape_supercell`](@ref), which is the
fundamental reshaping function. Rename `resize_periodically` to
[`resize_supercell`](@ref).

The constructor [`SpinInfo`](@ref) now requires a ``g``-factor or tensor as a
named argument.
```

### DIFF
--- a/docs/src/versions.md
+++ b/docs/src/versions.md
@@ -1,23 +1,21 @@
 # Version 0.5.0
 
-This version includes many **breaking changes**.
+**New features**.
 
-Added support for Dipole-mode Linear Spin Wave Theory. (Thanks Hao Zhang!)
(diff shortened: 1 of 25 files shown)


kbarros:
@ddahlbom and @Lazersmoke, there is not much to review in the implementation, but I added you as reviewers in case you have comments about these breaking changes.

kbarros:
David gave the thumbs up to merge.

Issue #110, opened 2023-07-29
Setting onsite coupling to zero doesn't remove it


Lazersmoke:
https://github.com/SunnySuite/Sunny.jl/blob/188846860b8fb6a447f8400f854311c340003b68/src/System/OnsiteCoupling.jl#L144

I use this in parameter sweeps, where `set_onsite_coupling!(sys, <zero matrix>, 1)` goes after `set_onsite_coupling!(sys, +0.1 * Sz^2, 1)` and should remove that coupling

kbarros:
Agreed.

PR #111, opened 2023-07-30
Allow `set_onsite_coupling!` with zero

kbarros:
This fixes https://github.com/SunnySuite/Sunny.jl/issues/110 and performs some additional reorganization.


kbarros:
The proposed new behavior is:

```julia
sys = System(...)
S = spin_operators(sys, 1)
set_onsite_coupling!(sys, zero(S[3]), 1) # does nothing
set_onsite_coupling!(sys, S[3]^2, 1)      # sets a coupling
set_onsite_coupling!(sys, zero(S[3]), 1) # emits a warning and removes coupling
set_onsite_coupling!(sys, S[3]^2, 1)      # sets a coupling *without warning* (Sunny will not detect onsite interaction)
```

PR #116, opened 2023-08-04
Wavevectors should be specified in absolute units.

kbarros:
This is a work in progress.

Provide `cryst.recipvecs` in analogy to `cryst.latvecs`. Using this, it becomes easy to manually convert from RLU to absolute units before calling Sunny functions.

The FeI2 tutorial has been updated. Averaging of 120 degree rotations becomes natural in absolute units.

Significant simplifications to LSWT become possible by working in absolute units everywhere. In particular, it is no longer necessary to keep track of an "original" crystal.

The tests involving SampledCorrelations and binning are failing. This part of the code needs to be updated to use absolute units. I think this will lead to nice simplifications. See some "FIXME" markers for a subset of things that should change. Functions like  `bin_absolute_units_as_rlu!` and `bin_rlu_as_absolute_units!` are no longer needed.

### DIFF
--- a/examples/fei2_tutorial.jl
+++ b/examples/fei2_tutorial.jl
@@ -246,19 +246,21 @@ plot_spins(sys_min; arrowlength=2.5, linewidth=0.75, arrowsize=1.5)
 swt = SpinWaveTheory(sys_min);
 
 # Next, select a sequence of wavevectors that will define a piecewise linear
-# path in reciprocal space. The coefficients of each $q$-vector are given in
-# reciprocal lattice units (RLU), i.e., are multiples of the recripocal lattice
-# vectors. The function [`connected_path`](@ref) will linearly sample between
-# the provided $q$-points with some `density`.
-points = [[0,   0, 0],
-          [1,   0, 0],
-          [0,   1, 0],
-          [1/2, 0, 0],
-          [0,   1, 0],
-          [0,   0, 0]] 
-labels = string.(points)
-density = 600
-path, markers = connected_path(swt, points, density);
+# interpolation in reciprocal lattice units (RLU). Sunny expects wavevectors in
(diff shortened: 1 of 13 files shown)


ddahlbom:
I realized this has very little impact on most of the sampled correlations material. I basically added one line of code at the beginning of `intensities_interpolated` to convert to the input to RLU with respect to the reshaped crystal, and I changed `all_exact_wave_vectors` so that it returns absolute units. 

I initially thought the change could lead to some simplifications, but that thought was premature. (Whatever the input vectors are, they have to be rounded to the nearest available wavevector. So the wave vectors, in RLU and absolute units, need to be re-calculated no matter what the format of the input.)

A couple comments:
1) Kipton suggested moving atom information into `SpinInfo`, which would sidestep some of the complexity of applying form factors. Probably it would be best to do this in another PR, possible the one bringing form factors into SWT calculations. I want to chat with Hao about that this next week.
2) This PR starts to break down the internal convention of having `q` refer to RLU and `k` refer to absolute units. That seems fine, but if we want some new convention, we should agree on something.

Lazersmoke:
Essentially no change is required to the binning code, since absolute units were already the default input to the `covectors` matrix.

# about `connected_bath_bins`:
There is some question about whether `connected_path_bins` should have a fixed density in RLU or in absolute units. Since the scattering vectors are uniformly spaced in RLU, it's more appropriate to space equally in RLU to get the maximum resolution. But at the same time, uniform spacing in absolute units is more physical in some sense. In short: RLU is optimal for Sunny data, whereas most experiment data will be in absolute.

Since `connected_path_bins` will not be called when using a `BinningParameters` derived from experiment data, it seems like it should be in RLU still! Perhaps it should accept the path in absolute units, though, but then the density parameter has an unnatural interpretation.

PR #119, opened 2023-08-09
FormFactor update

kbarros:
FormFactor no longer requires an atom index. Instead, the list of FormFactors must appear in the same order as the site symmetry "classes" of the crystal, i.e. as the `crystal.classes` field. The order of these classes is invariant to crystal reshaping.

FormFactors now work for LSWT.

David says this code is on the performance critical path for SampledCorrelations measurements, so we should do a benchmark before merging.

### DIFF
--- a/docs/src/versions.md
+++ b/docs/src/versions.md
@@ -55,6 +55,10 @@ fundamental reshaping function. Rename `resize_periodically` to
 The constructor [`SpinInfo`](@ref) now requires a ``g``-factor or tensor as a
 named argument.
 
+The constructor [`FormFactor`](@ref) no longer accepts an atom index. Instead,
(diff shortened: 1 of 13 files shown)


kbarros:
The unit tests on FormFactors are currently sparse. Maybe we should update "Sampled correlations reference" to include form factors before and after?

ddahlbom:
Since there's only one ion type, the form factor correction should appear as a simple mask in Q-space which we could generate independently of the calculation. We could add a test on an instant structure factor that tests whether application of the mask to the uncorrected results yields something that is approximately the same as what comes out of Sunny with the form factor correction turned on. (Perhaps this is what you meant.)

I can set this up in the next day or two if it's helpful.

ddahlbom (approved):
Looks great to me.

PR #121, opened 2023-08-11
Finite Temperature FeI2 Example

ddahlbom:
This is a work in progress. It is mostly a recycling of the second half of the old tutorial.


kbarros:
Looks great to me!

ddahlbom:
Looks great to me. Glad you worked in some more about using Langevin for optimization.

PR #122, opened 2023-08-11
Swt speedup

Lazersmoke:
Sorry for messy commit, you can ignore `scatter_bin_centers` and `intensities_bin_multisample` (they are for inverse problem and I meant to put them on main/separate PR)

This is intended to just make LSWT (much) faster without changing any behavior, but another set of eyes on this would be much appreciated to make sure I didn't accidentally change something important. The tests pass but I'm not sure if this code is covered by the tests or not!


kbarros:
Hi @Lazersmoke, thanks a lot for optimizing this. I think it's complicated enough that we should have a conversation including @Hao-Phys to discuss design.

Another low-hanging fruit: The dynamical matrix has a 2x2 block structure, which I think @Hao-Phys calls `[Hmat11 Hmat12; Hmat21 Hmat22]`. The upper and lower blocks are linked by `Hmat22 = conj(Hmat11)` and `Hmat21 = conj(Hmat12)`. So it might be best to only fill in the top two blocks in the loops, and set the bottom two blocks at the very end.

kbarros:
Let's include also @ddahlbom in the discussion if they're available.

Lazersmoke:
Harry found that it's impossible to use `:full` with spin wave theory due to goldstone mode check, which we were already planning to subsume into a different part of the code, so I've removed it here.

PR #123, opened 2023-08-12
Wavevectors back to RLU

kbarros:
This is a work in progress. Tests are passing, but I think @Lazersmoke needs to update binning code.

This won't be ready until the generated docs for `fei2_classical.jl` and `binning_tutorial.jl` look good.

Good news: The examples are now serving as nice tests!


kbarros:
Note that LSWT now uses `q_reshaped` internally, which can be our naming convention for RLU for the _reshaped_ supercell. This is what @Hao-Phys  requested.

kbarros:
- `fei2_tutorial.jl` is now giving the correct results.
- There are still scaling artifacts at the bottom of `fei2_classical.jl`. The fix may require an update to `connected_path_bins` to accept RLU.

Other updates I think we should include with this PR (help requested from @Lazersmoke  and @ddahlbom):

1. Adopt a consistent naming convention throughout Sunny to indicate the units of a wavevector. Previously @ddahlbom was distinguishing `q` and `k`, but I think that's too subtle. I have adopted `q` for a wavevector in original RLU, `q_reshaped` for a wavevector in RLU for the reshaped crystal, and `q_absolute` for a wavevector in absolute units. We need all three because: (a) LSWT and SampledCorrelations want to work with `q_reshaped` internally, (b) Sometimes we need `q_absolute`, e.g., for calculating the form factor, and (c) The user interface is entirely in `q`, i.e. the original RLU. 

2. I have renamed `connected_path` to `reciprocal_space_path`, which reflects the more specific nature of this function. We will need some analogous renaming of `connected_path_bins`.

PR #125, opened 2023-08-17
Optimizations to LSWT

kbarros:
Sam's LSWT optimizations.


src/SpinWaveTheory/SWTCalculations.jl
```diff
@@ -53,33 +64,68 @@ function swt_hamiltonian_SUN!(swt::SpinWaveTheory, q_reshaped::Vec3, Hmat::Matri
...
-            tTi_μ = s̃_mat[:, :, :, sub_i]
-            tTj_ν = s̃_mat[:, :, :, sub_j]
-            sub_i_M1, sub_j_M1 = sub_i - 1, sub_j - 1
+            # Fill dipole part of tTi_μ
+            tTi_μ[:, :, 1:3] .= view(s̃_mat,:, :, 1:3, sub_i)
+            tTj_ν[:, :, 1:3] .= view(s̃_mat,:, :, 1:3, sub_j)
+            
+            # Fill quadrupole part of tTi_μ
+            tTi_μ[:, :, 4:8] .= view(Q̃_mat,:, :, 1:5, sub_i)
+            tTj_ν[:, :, 4:8] .= view(Q̃_mat,:, :, 1:5, sub_j)
```
Lazersmoke, line 79: @Hao-Phys this the main thing that it would be good to check on; are these comments accurate? and what exactly are the `tTi_mu` and `t_mu_ii` variables meaning here?

Hao-Phys: These comments are correct. `tTi_mu` means the SU(N) generators in the local frame at site `i`, and `tTj_mu` means the same but at site `j`. Here sites `i` and `j` are connected by a bond.

kbarros:
Is it possible to make a complicated (and arbitrary) LSWT model that exercises _everything_ in a simple unit test at some arbitrary wavevector? This might be possible without much work if the wavevector is commensurate. Having that in place seems nicer prior to optimizations.

Lazersmoke:
I will say, as I'm making more changes and actually breaking things accidentally, the existing `test_lswt` appears to do a good job of catching the errors

kbarros:
Yeah the current tests are super helpful.

Specific things that would be nice to test: 

- multi-Q orderings that exercise the phase averaging more fully.
- the 3x3 component from `:full`

I think @hlane33 may be working on this?

hlane33:
This is correct @kbarros.

Hao-Phys:
Hi @Lazersmoke, thanks a lot for making lswt fast! Your changes look great to me. Glad to know that the current test_lswt is able to catch errors. For the more general example, I suggest that we should ask the experimentalists for the most complicated models they have in mind. Then we use spinW to check.

Hao-Phys:
Actually, if I recall correctly. FeI2's s(q,w) does have off-diagonal components. This model is quite complicated. We can use spinW to compute the SU(2) spectra and compare. (Potentially we can also add artificial DM interactions if they are symmetry-allowed)

kbarros:
Agreed that it's nice to reproduce various experimentally relevant calculations.

But for purposes of a unit test, an artificial model filled with the "kitchen sink" of features might serve best as a unit test:

- Crystal with multiple sublattices, and some complicated reshaping
- Multi-Q order
- Exchange, DM, high order onsite coupling, biquadratic coupling
- Measure all 3x3 elements S(q,omega), some complex.
- Measured at arbitrary q vector.

We can calculate the (:SUN and :dipole) intensities once, and then right a unit test to make sure these results never change as we update the LSWT code. We can also compare with the SpinW result in dipole mode (for this, it's possible we need to turn off biquadratic in SpinW if we don't trust the SpinW result).

Lazersmoke:
Since this is blocked on testing anyway, I went ahead and made some more renamings, simplifications, etc. Something appears to be subtly broken now, but I've tracked it down to having to do with the relation between specifically `Hmat11` and `Hmat22`. Are these not supposed to be complex conjugate transposes of each other? It seems that they are transposed, but only the phase `exp(iq . Delta R)` is conjugated. Ideally we would like to compute only `Hmat11`, and then conjugate to get `Hmat22`. Maybe @Hao-Phys can comment if this is possible?

Hao-Phys:
Hi @Lazersmoke . No the relation is not simply a transpose. The relation is give by Hmat22(k)=transpose(Hmat11(-k))

Lazersmoke:
It now passes tests and should be correct 🎉 ! Thank you!

Lazersmoke:
Before optimization:

image

After:

image

Lazersmoke:
I've just rebased this so that we're up-to-date on the `Requires.jl` change

kbarros:
This is still blocked on a "golden reference" test. Per our discussion today, the test doesn't need to be very physical, it just needs to verify that the behavior of the LSWT calculation will not change for any conceivable input. It sounded like @hlane33 plans to implement such a test, is that correct? Once we get that I am happy to merge this.

hlane33:
Yes I will put something together by the end of the weekend

Lazersmoke:
It's actually quite difficult to put everything at once into once test, but I've put everything you could imagine for an unphysical SU(N) test into a kitchen sink test, and verified that it also passes on `main`

kbarros:
We merged Sam's nice SU(N) "golden test" here: https://github.com/SunnySuite/Sunny.jl/commit/c8aec95061218229bb1b4db8d777d1a0e479c13b

The deviation between "terminal" and "test" environments @Lazersmoke mentioned above could be due to Julia configuration (@Lazersmoke tracked it down to the three-term `dot` function). Since the difference is only in the last digit, a solution is to weaken the test tolerance.

I cleaned up the Langasite test, and @hlane33 is welcome to review the new version.

I rebased on `main` as two commits, and I think we can merge them without squashing.

PR #126, opened 2023-08-17
Sundry interface tweaks for 0.5

ddahlbom:
- `ωs` -> `available_energies`
- `all_exact_wave_vectors` -> `available_wave_vectors`
- FeI2 classical tutorial now presents single-energy cuts without skew.
- No longer possible to define an `intensity_formula` without specifying a mode
- Default contraction mode is now `:trace`

I couldn't think of a nice way of making of forcing a user to specify a mode with every intensity call because of potential redundancy (or conflict) with an explicitly given formula. Put another way, the mode is really set in the formula. So, if we wanted to make sure all calls to an `intensity` function specify a mode explicitly, we would really have to require a formula (rather than a mode). I kind of like that idea, but it probably would make things too cumbersome for first time users. Let me know your thoughts if you don't like this arrangement.

Default `:trace` feels more natural to me (isn't overly biased toward people working with scattering data), but I don't think it's a hill I'm willing to die on.

@Lazersmoke You'll want to check that I've used the binning code in a reasonable way in the tutorial and make sure I haven't messed anything up when playing with the defaults.


examples/fei2_classical.jl
```diff
@@ -324,18 +325,17 @@ fig#hide
...
-
-params.binstart[4] = target_ω
-params.binend[4] = target_ω + 0.15
-params.binwidth[4] = 0.3
+params = BinningParameters(
+    [-3,-3, 0, target_ω],           # Bin start (h,k,l,ω)
+    [3, 3, 0.15, target_ω + 0.15],  # Bin end
+    [.2, .2, .3, 0.3]               # Bin width
+)
+Sunny.bin_rlu_as_absolute_units!(params, cryst.recipvecs)
```
ddahlbom, line 338: It seems to me that this should be called `bin_absolute_units_as_rlu!`. The bins are specified in absolute units (used default orthogonal coordinates), and I'm trying to convert them to RLU. Let me know if I am confused.

kbarros: Ugh, yeah, I see how absolute is a lot nicer here. Should we do a short call to brainstorm the ideal user interface?

kbarros: Given that `bin_rlu_as_absolute_units!` is for now private, I think we can merge this PR and fix up later.

Lazersmoke: (Oh and I should say, using the `unit_resolution` stuff makes the x and y axes as high resolution as possible given the lattice size at essentially zero additional computational cost)

ddahlbom: Thanks for the tips. I suspected I was abusing the functionality. My goal was just to get the figure to look like this,

image

i.e., to remove the skew. In other words, I wanted to define regular bins in absolute units directly. I get this is non-ideal from a certain perspective, but I think this is the way one would want to visualize these data.

Your second approach will work, but if I don't fudge the bin widths, there will be gaps. So, for example, the following gets the desired effect.
```
params = unit_resolution_binning_parameters(sc)
params.covectors[1:3,1:3] .= cryst.recipvecs
omega_width = 0.3
params.binstart[4] = target_ω
params.binend[4] = target_ω + (sc.Δω/2)
params.binwidth[4] = omega_width
params.binwidth[1:2] .*= 2

params.binstart[1], params.binstart[2] = -3, -3
params.binend[1], params.binend[2] = 3, 3

integrate_axes!(params, axes = [3,4])
```

image

Lazersmoke: Oh darn yeah I see. The root cause is the eigenvalues of recipvecs are not all one or whatever. Maybe try multiplying the binwidth by `1.64` instead? (Close to largest eigenvalue of recipvecs)

kbarros: Apart from the questions about `unit_resolution_binning_parameters`, this looks great to me! You have my thumbs up after convergence with @Lazersmoke re. binning interface questions.

ddahlbom: I find the minimal number necessary for rescaling the binwidths (to avoid empty bins) is about 1.78. It might be aesthetically nicer with a lower resolution, but I'll leave it at that. This can always be changed later.

ddahlbom:
Ok, a formula is now required for all `intensities` functions. 

I think I am happy with the current state of this. Let me know if there are any final objections.

PR #129, opened 2023-08-18
Switch from Requires.jl to package extensions

kbarros:
Remove Requires.jl in favor of Julia 1.9 "package extensions". Two motivations are (1) it becomes possible to specify `[compat]` entries for weak dependencies (e.g., we can enforce recent versions of GLMakie), and (2) it becomes possible to use PrecompileTools.jl which I'm hoping can significantly reduce TTFX in the common cases.

I hit a few snags in making this PR:

- Unlike Revise.jl, package extensions can only add methods to existing functions, they cannot create new functions in the base package. The workaround seems to be defining "stub" functions in the original Sunny module, and adding methods in the extension packages.
- It gets inconvenient to have to define a new stub for each private function. I used a hack to create a `Sunny.Plotting` module which will store the private plotting functions.
- The Documenter.jl `@autodocs` feature is not finding doc strings for functions defined through package extensions. https://github.com/JuliaDocs/Documenter.jl/issues/2124#issuecomment-1683268037. The workaround seems to be creating `@docs` blocks that list each extension function explicitly.

This PR also makes `DynamicPolynomials` a required dependency of `Sunny` again. That choice increase the time for `using Sunny` from about `1.15s` to `1.3s`-ish, and most of that increase comes directly from `using DynamicPolynomials`. I think we need to do it though (or roll our own dynamic polynomials) because this is the most natural way to specify classical polynomials of commuting spin variables. Maybe we can win back load-time efficiency in the future by, e.g., adding precompile statements.

I think it's ready to merge if @ddahlbom  gives a thumbs up.


docs/make.jl
```diff
@@ -38,7 +38,7 @@ Documenter.makedocs(
     format = Documenter.HTML(
         prettyurls = get(ENV, "CI", nothing) == "true",
         ansicolor = true
-    );
+    ),
     draft
```
Lazersmoke, line 42: Does this not break??

ddahlbom:
I have to admit I'm a bit confused about the hack you used @kbarros. Perhaps we can chat about it at some point. But I everything looks good to the degree that I understand it.

@Lazersmoke Is your primary concern about users in some situation where they can't display unicode? How common do you think that will be? (I guess remote work on a terminal could be a problem.)

I can't think of a more satisfying alternative at the moment. From an interface standpoint, the symbols are kind of "magical" and unique in the interface, so it makes sense that they'd be the only example of unicode that Sunny exposes. I don't think we want to take over regular `S` and `O`.

We could provide a non-unicode synonym that's uglier, but usable in a basic terminal. Maybe `Ss` or `Os`, to emphasize that they're vectors.

In any case, from my perspective it's good to merge.

docs/make.jl
kbarros: It works because semicolon has been moved to the top of the keyword argument list.

kbarros:
> Still not sure about the Unicode here

The Unicode is being removed from the public interface. Maybe you are noticing that I removed the links to the historical symbols in versions.md? But this PR actually removes those symbols from the public API.

test/test_operators.jl
```diff
@@ -241,13 +241,16 @@ end
 @testitem "Symbolics" begin
     import IOCapture, DynamicPolynomials
 
+    𝒪 = Sunny.𝒪
+    𝒮 = Sunny.𝒮
```
Lazersmoke, line 245: @kbarros is the idea that these are not public so it's ok? Confused why these lines are added here

src/Sunny.jl
```diff
@@ -127,24 +127,12 @@ include("MonteCarlo/WangLandau.jl")
...
-    # Importing WriteVTK will enable saving files to view with ParaView
-    @require WriteVTK="64499a7a-5c06-52f2-abe2-ccb03c286192" begin
-        include("VTKExport.jl")
-        export export_vtk
-    end
-
-    # Importing DynamicPolynomials will enable certain symbolic analysis
-    @require DynamicPolynomials="7c1d4256-1411-5781-91ec-d7bc3513ac07" begin
-        include("Operators/Symbolic.jl")
-        export 𝒪, 𝒮, print_classical_stevens_expansion, print_classical_spin_polynomial
```
kbarros, line 146: Deletion of Unicode API.

test/test_operators.jl
kbarros: Yeah these were removed from the public API and will not be exposed to the users. More cleanups to follow. This PR is focused on switching to package extensions.

ddahlbom (approved):
Thanks for the clarifications. It makes sense to me now. Looks good to me!

PR #130, opened 2023-08-18
f in intensity_formula expects absolute units

kbarros:
@Lazersmoke: Help requested to clarify which units go where. This particular change looks correct given our previous discussions, but I think we need to double check the `q` usage elsewhere.


kbarros:
1. The `calc_intensity` defined in SWTCalculations.jl, which is modified by this PR, seems to take as input `q` in RLU. I think this is what everyone is requesting.
2. The `calc_intensity` defined in for `SampledCorrelations` seems to take as input `q_absolute` instead. See: https://github.com/SunnySuite/Sunny.jl/blob/main/src/SampledCorrelations/DataRetrieval.jl#L82

Is this deviation intentional? cc. @ddahlbom .

Lazersmoke:
What needs to be checked is that the polarization correction is computed using q_absolute. If I'm reading it correctly that's what this PR does so looks like it was incorrect before 😅 thanks for the fix

PR #143, opened 2023-08-24
Stable bogoliubov transform

Lazersmoke:
This modifies the algorithm used to compute the bogoliubov transform very minorly. It sets the sign of the real part of the first entry of each eigenvector to be positive so that calling `bogoliubov!` on nearby hamiltonians results in nearby (or at least near-er than before) bogoliubov transform matrices. Is this allowed @Hao-Phys ? Can the eigenvectors be rotated by arbitrary phases, or just multiplied by -1? It may be even more stable to rotate them to have first entry real rather than simply positive real part.

This is in preparation for differentiating through `bogoliubov!` in some clever way.


Hao-Phys:
Hi @Lazersmoke. If I understand correctly, you are trying to fix the U(1) gauge for the eigenvectors. Yes, in principle, the phase should be arbitrary. However, I am not sure about your criterion for fixing the gauge.

kbarros:
Let's postpone this until we reach a conclusion about the strategies for gradient calculation.

Issue #146, opened 2023-08-28
Multithreading of sampled correlations


kbarros:
Create a variant of `add_sample!` that works in parallel using all available Julia threads.

PR #162, opened 2023-09-06
Parallelism examples

ddahlbom:
This is a work in progress. It is just a tutorial with some simple patterns for achieving parallelism. It also includes a new function `merge_correlations`. 

I'm still a bit unsure how to handle the presentation of code that we may not execute consistently (since we aren't guaranteed of any particular resources in the github CI). Possibly we don't execute at all and just present as text and code snippets.


Lazersmoke:
Since the actual outputs of this example aren't really of interest, there's no reason to really execute them during the docs build. I think there was the same situation with the WriteVTK example

ddahlbom:
Yes, that sounds sensible.

ddahlbom:
Since we decided to just provide examples rather than rework internals, for this

Closes #146

PR #167, opened 2023-09-19
General Observables 2

kbarros:
This rebases and squashes the commits from https://github.com/SunnySuite/Sunny.jl/pull/156. I also fixed some errors and cleaned up some style. Let's use this format for future PRs (a minimal number of squashed PRs on top of main). For PRs, always use `git rebase` and not `git merge` to make it easier to read the changes.


kbarros:
@Lazersmoke can you please take a look at the failing Pyrochlore test at the bottom of `test_lswt.jl`. It is just the intensities check that is failing, and this is due to a shape error in the array comparison.
https://github.com/SunnySuite/Sunny.jl/blob/e4b261a1407fafede82564c242734fdb972a1595/test/test_lswt.jl#L464

test/test_lswt.jl
```diff
@@ -60,7 +60,10 @@
     # to one arbitrary k-vector, and round to 10 digits of accuracy.
     formula = intensity_formula(swt,:full,kernel = delta_function_kernel,formfactors = [FormFactor("Fe2")])
     _, is_full = intensities_bands(swt,[ks[1]],formula)
-    is_full_flattened = reinterpret(reshape, ComplexF64, is_full)[:]
+    is_full_flattened = zeros(ComplexF64,720)
+    for i = 1:80
+        is_full_flattened[9*(i-1) .+ (1:9)] .= is_full[i][:]
+    end
```
kbarros, line 66: Why are these additional lines needed?

test/test_lswt.jl
```diff
@@ -389,7 +393,11 @@ end
     
     @test isapprox(disp[:], reverse(SpinW_energies); atol=1e-3)
     
-    intensities_reshaped = reinterpret(reshape, ComplexF64, intensities)[:]
+    intensities_reshaped = zeros(ComplexF64,size(SpinW_intensities))
+    for i = 1:(length(SpinW_intensities)÷9)
+        intensities_reshaped[9*(i-1) .+ (1:9)] .= intensities[i][:]
+    end
+
```
kbarros, line 400: Here too.

kbarros:
General comment: Does this PR introduce a user interface change? If so, where can I find examples of specifying general observables? Are there examples included in the tests?

kbarros:
General comment: The spacing conventions deviate from the [BlueStyle](https://github.com/invenia/BlueStyle). For example, please use this: `f(x, y; z=2)`. Once all the outstanding PRs get merged, we can consider setting up [JuliaFormatter](https://github.com/domluna/JuliaFormatter.jl) to run as a Github action.

test/test_lswt.jl
Lazersmoke: Previously, this read `is_full_flattened = reinterpret(reshape, ComplexF64, is_full)[:]`, which fails because `SMatrix{3,3,ComplexF64}` isn't `isbits`. I'm not sure how this ever worked in the first place, but this code simply converts the intensities produced by the code to the shape of the golden reference intensities.

Perhaps a `Matrix` should be returned instead of `SMatrix`? Would that improve things?

Lazersmoke:
> General comment: Does this PR introduce a user interface change? If so, where can I find examples of specifying general observables? Are there examples included in the tests?

The whole UI change is this:

https://github.com/SunnySuite/Sunny.jl/blob/e4b261a1407fafede82564c242734fdb972a1595/src/SpinWaveTheory/SpinWaveTheory.jl#L37

This constructor doesn't have a doc string, but the `struct` does. Is that legal? In any case this isn't documented or tested[1] for `SpinWaveTheory` yet. But the docs are already duplicated for `dynamical_correlations` and `instant_correlations`: https://github.com/SunnySuite/Sunny.jl/blob/e4b261a1407fafede82564c242734fdb972a1595/src/SampledCorrelations/SampledCorrelations.jl#L121-L131

so we should find a way to de-duplicate this, ideally. One option is to document and/or export `parse_observables` (which is the shared internal function which parses the multiple allowed formats of the `observables` and `correlations` kwargs into an `ObservableInfo`) https://github.com/SunnySuite/Sunny.jl/blob/e4b261a1407fafede82564c242734fdb972a1595/src/Operators/Observables.jl#L31

[1] Re: testing. The (default) dipole observables are simply custom observables, so in some sense the whole thing is already being tested, as long as we believe that the dipole observables aren't special enough to break things. Perhaps a test case could use a non-hermitian observable, which is a case not covered by dipole observables.

Lazersmoke:
In case it's not clear, the supported kwarg inputs to the internal function `parse_observables` are:

Allowed formats for `observables`:
- `observables = [randn(3,3), randn(3,3), ...]` gets named `:A`, `:B`, ...
- `observables = [:obsOne => randn(3,3), :obsTwo => randn(3,3), ...]` for custom names (recommended)
- `observables = nothing` (default) yields `:Sx`, `:Sy`, `:Sz` dipole observables

Allowed formats for `correlations`:
- `correlations = [(1,1),(3,4),...]` yields correlations `(:A,:A)`, `(:C,:D)`, ...
- `correlations = [(:A,:A),(:C,:D),...]` yields correlations `(:A,:A)`, `(:C,:D)`, ...
- `correlations = nothing` (default) yields all correlations (upper diagonal)

test/test_lswt.jl
kbarros: That's strange, I see this at the terminal:
```julia
m = randn(Sunny.Mat3)
@assert isbits(m)
```
I will think about it tomorrow and see if there's some reinterpret/reshape that's possible.

kbarros:
Thanks for fixing the bug and clarifying the UI. I will think about the docs question too.

ddahlbom:
> One option is to document and/or export `parse_observables` (which is the shared internal function which parses the multiple allowed formats of the `observables` and `correlations` kwargs into an `ObservableInfo`)

I suppose `parse_observables` could just be changed into the constructor for `ObservableInfo`, and the user would be expected to make an `ObservableInfo` just as they make a formula. But that does make a public interface commitment, which we may wish to shy away from at the moment.

I get the point about dipoles no longer being a special case, so the default option exercises the new approach automatically. I'll try to think if there's something else we can do to exercise this.

I think committing this soon in its current form makes sense. It consistently and nicely extends the present behavior of SampledCorrelations to LSWT. Maybe we just accept some doc duplication for `SpinWaveTheory` at the moment. We can have a discussion before making any more substantial interface commitments.

test/test_lswt.jl
kbarros: I agree there's something strange here. Let's see what they say here: https://github.com/JuliaArrays/StaticArrays.jl/issues/1201

kbarros:
> I suppose parse_observables could just be changed into the constructor for ObservableInfo

I like this idea -- in our eventual final interface, the user can construct their own `ObservableInfo` and then pass it pre-formed to the various intensity formula functions. But given the current state of flux, we can hold off on that change for now.

Here is another question that is more urgent: Am I understanding correctly that with `:full` the `intensities_bands` function now returns a list of `SMatrix{3,3}` rather than a flattened `Vector` of 9 elements? This seems to be the origin of the breakage in `test_lswt.jl`, and will also break user-code that happens to be using `:full` (did we include anything like this in our examples?). Is this a breaking change we are motivated to make, and do we want to make it now? What are the tradeoffs?

- If we want to make this breaking change, then we at least need to add documentation in `versions.md`.
- I am thinking maybe we _don't_ want to make this breaking change. My long term goal is to remove all dynamic typing behavior from the `intensities*` functions (except maybe some very simple conversions in the constructors). It's a simple story to say: The 3x3 matrix of intensities should be "just another list of observable pairs".

kbarros:
Regarding the reshaping mystery, the mistake was forgetting to include the final fourth argument (`NObs2`) when defining `Contraction{SMatrix{NObs, NObs, ComplexF64, NObs2}}`. With the latest commit,   https://github.com/SunnySuite/Sunny.jl/pull/167/commits/d65c8020baec5fdf6ba9e1879d5c2f9929a1e8d8, it is fixed. Now the reshapings in `test_lswt.jl` work again.

The final question to address: does this introduce a breaking change, and do we want to do that? Need input from both @Lazersmoke and @ddahlbom, maybe on a short call.

Lazersmoke:
The previous behavior was to dump whatever correlations were stored in `sc.data` on the user with little explanation. They ended up getting a six element vector from SampledCorrelations, or a 9 element vector for SWT, neither of which really has a canonical ordering, especially in the presence of general observables. I recently sent someone this line of code from deep within sunny to explain what they were getting from `:full` (in v0.4.3): 

https://github.com/SunnySuite/Sunny.jl/blob/370d8016c3ac6fcd8387e59b6b02a9a0583eaf8b/src/StructureFactors/ElementContraction.jl#L46

The correlation matrix (3x3 by default) is what users really want to interact with when they ask for the full tensor

ddahlbom:
I agree with Sam that the previous behavior was close to uninterpretable, and putting it into 3x3 matrix is the most common thing a person would want from a `:full` call. But I also agree that it's sort of a special case that doesn't sit well with the current set of abstractions: every other "mode" will give a number, and in principle `:full` should return some kind of sparse matrix, the size and interpretation of which depends on the observable information given. 

The `:full` approach is also conceptually more fundamental, i.e. it is what exists before a contraction and should correspond in some sense to a "no op" contraction.

I think we may want to move away from the current `Contractor` approach altogether. Conceptually (not necessarily from an interface standpoint), there should be a layer of functions that return the "full" results. There should probably be functions that take these results and the observable info and produce some correlation matrix that's user friendly. And then there should be contraction function (probably not as a type type).

The trickiness lies in how to resolve this in a way that is both performant and presents a reasonable user interface. 

A lot of decisions for the original SampledCorrelations intensities code were based on the notion that it might be plugged into an interactive plotting interface, so a lot of effort was made to pool many steps together without allocations. I'm not sure this is really a priority any more. We can probably do something that has multiple clearly-separated stages of processing and accept allocations between these stages.

PR #173, opened 2023-09-25
Updates to docs build

kbarros:
Still a work in progress.


kbarros:
- Significantly reduce dependencies in docs/Project.toml
- Use unicode subscripts and superscripts instead of embedded Latex, where possible.
- Include SpinW ports with main docs.
- Remove one_dim_chain.jl from main docs. It is failing sporadically, and might fit better in SunnyContributed anyway.
- Link to local assets where possible.
- General cleanups.

Issue #63, opened 2023-02-17
Support biquadratic interactions for SU(N) coherent states, and beyond


kbarros:
Currently we support scalar biquadratic interactions in `:dipole` mode. Eventually we should allow this also for `:SUN` mode. The latter would traditionally require thinking about a basis for SU(N) generators that explicitly defines spin operators $\hat S$ as well as operators $\hat Q$ quadratic in spin.

Rather than hand-coding these generators, it might be better to work entirely with polynomials of the spin operators, and their tensor products, e.g. $S_j = S \otimes I$ and $S_k = I \otimes S$ to represent spins for sites $j$ and $k$ in the context of the bond $(j,k)$. This approach would generalize to any order exchange interaction, as well as to entangled units.

kbarros:
Here is some code to construct the spin operators as tensor products, and to use SVD to compress operators.

```julia

using Sunny, LinearAlgebra

# Produces a matrix representation of a tensor product of operators, C = A⊗B.
# Like built-in `kron` but with permutation. Returns C_{acbd} = A_{ab} B_{cd}.
function kron_operator(A::AbstractMatrix, B::AbstractMatrix)
    TS = promote_type(eltype(A), eltype(B))
    C = zeros(TS, size(A,1), size(B,1), size(A,2), size(B,2))
    for ci in CartesianIndices(C)
        a, c, b, d = Tuple(ci)
        C[ci] = A[a,b] * B[c,d]
    end
    return reshape(C, size(A,1)*size(B,1), size(A,2)*size(B,2))
end

# Use SVD to find the decomposition D = ∑ₖ Aₖ ⊗ Bₖ, where Aₖ is N₁×N₁ and Bₖ is
# N₂×N₂. Returns the list of matrices [(A₁, B₁), (A₂, B₂), ...].
function svd_tensor_expansion(D::Matrix{T}, N1, N2) where T
    @assert size(D, 1) == size(D, 2) == N1*N2
    D = permutedims(reshape(D, N1, N2, N1, N2), (1,3,2,4))
    (; S, U, V) = svd(reshape(D, N1*N1, N2*N2))
    ret = []
    for (k, σ) in enumerate(S)
        if abs(σ) > 1e-12
            u = reshape(U[:, k], N1, N1)
            v = reshape(V[:, k], N2, N2)
            push!(ret, (σ*u, conj(v)))
        end
    end
    return ret
    # test = U * diagm(S) * V'
[...]

PR #176, opened 2023-10-04
Generalized pair couplings

kbarros:
This PR provides initial support for https://github.com/SunnySuite/Sunny.jl/issues/63, with some differences:

1. It uses the built-in `LinearAlgebra.kron` as the tensor product operation.
2. Given a general interaction of the form $\sum_i A_i \otimes B_i$, this PR will "extract" the bilinear spin-spin exchange, and the scalar biquadratic parts, if present. This is done to reduce the number of terms present in the sum of tensor products, which is empirically the slowest part of a simulation.
3. The generalized pair couplings work in the classical SU(_N_) dynamics but not yet spin wave theory. This will be addressed in a future PR.
4. Currently `:dipole` mode simulations cannot handle anything but scalar biquadratic. That will be addressed in another PR.


Lazersmoke:
How many terms in the SVD sum do you need in the best/worst/typical case?

kbarros:
This illustrates the behavior on the this branch. The user interface is subject to change, and that's the current blocker for this PR.

```julia
using Sunny

S1 = spin_matrices(; N=3)
S2 = spin_matrices(; N=4)
Si, Sj = Sunny.to_product_space(S1, S2)

D = Si'*randn(3,3)*Sj
@assert length(Sunny.svd_tensor_expansion(D, 3, 4)) == 3

D = (Si'*randn(3,3)*Sj)^2
@assert length(Sunny.svd_tensor_expansion(D, 3, 4)) == 9

# In practice, though, parameters to `set_pair_coupling!` will pass through this
# decomposition function:
bilin, tensordec = Sunny.decompose_general_coupling(D, S1, S2; fast=true)
@assert length(tensordec.data) == 6  # Just 6 terms left after extracting 3x3 dipole-dipole part

# As an optimization, `decompose_general_coupling`, could also extract the 5x5
# quadrupole-quadrupole part. For this particular example, there would be no terms
# left in the "genereal" expansion. (Not yet implemented.)

# bilin, biquad, tensordec = Sunny.decompose_general_coupling(D, S1, S2; fast=true)
```

PR #178, opened 2023-10-08
Work towards entangled units

kbarros:


Lazersmoke:
@kbarros there's something funky with the system show instance in this PR. You should check that typing both `sys` and `[sys]` in the repl produce sensible output (the second one is used frequently for inverse modelling + distributed)

kbarros:
> @kbarros there's something funky with the system show instance in this PR. 

Extracted from this PR and fixed on main: https://github.com/SunnySuite/Sunny.jl/commit/920e4b7b57d304c0790a7d64326e96cf44368830

ddahlbom:
Recent pushes include:

1) Generalized interactions for SU(N) LSWT
2) Biquadratic functionality now uses the Stevens operators and works with a general 5x5 quadrupolar "exchange" matrix. (This needs a test of some sort.)
2) Some reorganization of the SU(N) LSWT code, mostly in the form of breaking things into discrete functions
3) Removal of extra buffers and some code simplifications
4) Elimination of all allocations from the construction of the SWT Hamiltonian
5) A few other performance enhancements (e.g., checking whether `bilin` or `biquad` is zero before executing associated code -- this is quite significant)
6) Construction of external magnetic field operators when the `SpinWaveTheory` is first constructed.
7) Addition of two tests:
    - Consistency of Hamiltonian construction when `extract_parts` is true and false
    - A reference test for spin ladder intensities

Further simplifications are possible -- we only really need the basic logic of  `swt_onsite_coupling!` and `swt_general_couplings!` and could eliminate both `swt_bilinear` and `swt_biquadratic`. I am leaving these for now, however, since we should discuss the implications for KPM. I left the contents and style of `swt_bilinear` and `swt_biquadratic` very similar to the original code so hopefully it is still readable to Hao, with whom I will discuss further. We may want to redo the indexing a bit once everyone is on the same page.

ddahlbom:
I reorganized the SWT file structure a bit. I did not end up doing much to the dipole part of the SWT code. Probably it should be kept in familiar form for the time being to make it easier for Hao to add the general biquadratic. This can be addressed in a future PR.

There's still a fair amount on the table as far as simplifying and accelerating the SWT code.  But I think what's done here represents a self-contained revision that incorporates the general interactions into SU(N) SWT.

As I mentioned on Slack, dipole and SU(N) modes seem to use different sign conventions for the external field. I'll examine this a bit. Once that's resolved, this will be good to go from my perspective.

kbarros:
> As I mentioned on Slack, dipole and SU(N) modes seem to use different sign conventions for the external field.

@ddahlbom mentioned elsewhere that the sign conventions are the same.

PR #180, opened 2023-10-24
Eigenmode viewer hooks

Lazersmoke:
This includes all helper functions and hooks needed to make the eigenmode viewer work.

### DIFF
--- a/docs/src/versions.md
+++ b/docs/src/versions.md
@@ -2,6 +2,9 @@
 
 ## v0.5.6
 
+* Animated spin dynamics is now possible. Call `notify` on the result of
+  [`plot_spins`](@ref) to trigger redrawing of the frame. The argument `colorfn`
+  to `plot_spins` supports animation of colors.
 * General pair couplings are now supported in [`set_pair_coupling!`](@ref) and
   [`set_pair_coupling_at!`](@ref). `:SUN` supports interactions of any order,
   but `:dipole` mode is limited to bilinear and biquadratic coupling of the
--- a/examples/extra/heisenberg_animation.jl
+++ b/examples/extra/heisenberg_animation.jl
@@ -1,0 +1,25 @@
+using Sunny, GLMakie
+
+# Square lattice
+latvecs = lattice_vectors(1, 1, 10, 90, 90, 90)
+cryst = Crystal(latvecs, [[0,0,0]])
+
+# Simple Heisenberg model
+sys = System(cryst, (10,10,1), [SpinInfo(1, S=1, g=2)], :dipole)
+J = -1.0
+set_exchange!(sys, J, Bond(1, 1, (1, 0, 0)))
+randomize_spins!(sys)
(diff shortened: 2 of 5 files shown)


Lazersmoke:
(since this is getting merged first, we'll see how the diff between `main` and `asymmetric` looks afterwards and proceed from there)

(also note this includes `plot_band_intensities` which is used in un-exported form by the eigenmode viewer)

kbarros:
I pushed some updates:
- Now colors can also be animated using `colorfn` argument.
- The `dipoles` to be animated can no longer be detached from `sys.dipoles`. To support that use case, use `clone_system`. The memory costs will probably be much less than all the GLMakie objects anyway.
- The result of `plot_spins` is an object that supports both `display` and `notify`. This way users don't see observables.

Why not expose Makie `Observables`? Ultimately, that interface is not a nice abstraction for complicated objects with interior mutability. Later, for example, `notify(fig)` might draw other things besides the dipoles and colors. I tried to make `color` an Observable following Makie conventions, but the code ended up being much less clean (both on the Sunny side, and the user interface side). I think the final result here is very nice to use. See, e.g., the new Heisenberg animation example: https://github.com/Lazersmoke/Sunny.jl/blob/eigenmode-viewer-hooks/examples/extra/heisenberg_animation.jl

Lazersmoke:
This will be very slow because it empties the whole array of points and dipole vectors and then rebuilds it on every single frame. The point of having `spin_data` be separate is that it is fast to redraw (and modular to notify, unlike `NotifiableFigure`).

If this is the way that you want this in main Sunny, that's fine, and I'll write my own viewer function in SunnyContributed :)

kbarros:
To the contrary, this is designed for maximal speed. Ping me on slack and
we can work out how your use case will fit.

kbarros:
Following discussion on Slack and various benchmarks, it appears that this implementation is getting about maximal speed. In particular, Makie is intelligent in ignoring extra notifications if the data hasn't changed. That is, these extra notifications don't matter.

```julia
notify.((vecs, pts, pts_shifted, arrowcolor))
```

For a system with 10k spins, I get these results for one notify

```
[COLORS ON]  0.012429 seconds (141.15 k allocations: 6.590 MiB)
[COLORS OFF] 0.009867 seconds (81.16 k allocations: 4.759 MiB)
```

This is much slower than I like, but I think the slowdown is fundamentally on the Makie side. Note the speedup we get if the `arrowcolor` data is unchanged, even though we spuriously `notify(arrowcolor)`.

Another important thing to know is that this pattern is non-allocating:

```julia
function f(a)
    l = length(a)
    empty!(a)
    for i in 1:l
        push!(a, 42)
    end
end

r = randn(10)

using BenchmarkTools
@btime f(r)
```

In particular, `empty!` will not actually free the underlying data -- it will be saved for future use. To actually free, you would need something like `size_hint!`.

PR #183, opened 2023-10-31
Better binning semantics

Lazersmoke:
per slack discussion

This is a breaking change + someone else should independently verify the correctness

### DIFF
--- a/src/Intensities/Binning.jl
+++ b/src/Intensities/Binning.jl
@@ -1,7 +1,7 @@
 
 """
-    BinningParameters(binstart,binend,binwidth;covectors = I(4))
-    BinningParameters(binstart,binend;numbins,covectors = I(4))
+    BinningParameters(binstartcenter,binend,binwidth;covectors = I(4))
+    BinningParameters(binstartcenter,binend;numbins,covectors = I(4))
 
 Describes a 4D parallelepided histogram in a format compatible with experimental Inelasitic Neutron Scattering data.
 See [`generate_mantid_script_from_binning_parameters`](@ref) to convert [`BinningParameters`](@ref) to a format understandable by the [Mantid software](https://www.mantidproject.org/), or [`load_nxs`](@ref) to load [`BinningParameters`](@ref) from a Mantid `.nxs` file.
@@ -12,7 +12,7 @@ Since the default `covectors` matrix is the identity matrix, the default axes ar
 `(qx,qy,qz,ω)` in absolute units.
 
 The convention for the binning scheme is that:
-- The left edge of the first bin starts at `binstart`
+- The first bin is centered at `binstartcenter`
 - The bin width is `binwidth`
 - The last bin contains `binend`
 - There are no "partial bins;" the last bin may contain values greater than `binend`. C.f. [`count_bins`](@ref).
@@ -20,10 +20,10 @@ The convention for the binning scheme is that:
 A `value` can be binned by computing its bin index:
 
     coords = covectors * value
-    bin_ix = 1 .+ floor.(Int64,(coords .- binstart) ./ binwidth)
+    bin_ix = 1 .+ floor.(Int64,0.5 + (coords .- binstartcenter) ./ binwidth)
 """
 mutable struct BinningParameters
-    binstart::MVector{4,Float64}
+    binstartcenter::MVector{4,Float64}
     binend::MVector{4,Float64}
     binwidth::MVector{4,Float64}
     covectors::MMatrix{4,4,Float64}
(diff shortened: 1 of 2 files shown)


kbarros:
Thanks. I don't have time/expertise to review carefully, but I think the idea makes sense. When comparing with experimental data, is it the case that there will already be well defined bin centers that we want to match? To make the golden references pass, is it possible to do some "shifting" of bins? (OK if that's out of scope, just curious, because it might overlap with the need to shift for comparison with experimental data.)

Lazersmoke:
Closing this for now because it's substantially outdated.

But recording here the original purpose of this: We want to be able to write down [-Inf,Inf] as a valid integration range of a histogram axis. The best-defined, least-special-cases way to do this is to record the center of the starting bin (zero in the `[-Inf,Inf]` case) instead of the left edge.

The main application of this +-Inf range is for the energy axis of `unit_resolution_binning_parameters(instant_correlations(sys))`. Currently, this is `[-1,1]` or some other such fake thing. In theory, this is a problem if you binned experiment data according to this prescription; but this never happens in practice, since experiment data comes with its own, better, binning parameters.

[N.B.: Even if you exported the `[-Inf,Inf]` range e.g. as a Shiver script, your results would *still* be wrong due to sort-of-"[binning effect](https://github.com/SunnySuite/SunnyContributed/blob/main/initiatives/Sam-GT/docs/bin_effect_tutorial.md)" arising from the spectrometer only having a finite incident energy. So this [-1,1] vs +-Inf range is really purely academic in nearly all contexts]

kbarros:
Thanks for documenting this Sam. I'm cc'ing @bingli621 because I wonder if questions of binning simultaneously tie in with questions of instrument resolution.

Lazersmoke:
Hi @bingli621 thanks for the comments! A question about the neutron kinematics you mentioned:

Typically we use DGS data that has been exported from Mantid and contains `NaN` values in the kinematically disallowed regions. See the transparent region in the first figure [here](https://github.com/SunnySuite/SunnyContributed/blob/main/initiatives/Sam-GT/docs/bin_effect_tutorial.md). The transparent region gets larger with increasing transverse bin size because I only keep the pixels which are kinematically allowed over the *entire* transverse integration range. Then, only the remaining non-transparent part is used for fitting.

The question: is this hard cutoff sufficient to capture the kinematic limits, or is there a more-sophisticated smoother version of the kinematic constraint that should be used instead? It feels like the pixels near the edge of the kinematic cutoff should be discounted...

bingli621:
Hi Sam, what you are doing is correct, but you are throwing away usable data in your fitting process. The way I do it is slightly different. You could cut more than one slice in the transverse direction, read out the NaN values and use them as a mask for your simulation. For example, if in the transverse direction, you are binning from -0.15 to +0.15, you could make 3 slices with the bin range [-0.15, -0.05], [-0.05, 0.05], [0.05, 0.15]. Each slice will have slightly different (q,E) coverage. When you fit, also simulate 3 slices, and apply the masks accordingly. Then you sum the 3 data slices, and the masked simulation slices, then perform the fitting. This way, when you sum, you don't need to throw away the data when NaN is in the range. All you need is to keep track of the NaN and normalize accordingly. 

I do my analysis this way because the summed data have larger coverage and look nicer though not exactly accurate... 
This NaN business also pops up in ARCS data where there are gaps between detector banks.
Let me know what you think about it.

Lazersmoke:
OK yes thanks! That all sounds good. About this part:

> Then you sum the 3 data slices, and the masked simulation slices, then perform the fitting. This way, when you sum, you don't need to throw away the data when NaN is in the range. All you need is to keep track of the NaN and normalize accordingly.

You can also just fit on the 3 slices individually (*not* summed), or indeed on the whole 4D histogram :) As long as you don't need to visualize the summed slice, this gives you best of both worlds.

> I do my analysis this way because the summed data have larger coverage and look nicer though not exactly accurate...

If there's a transverse dispersion, you can get artifacts that are even visually incorrect when you sum the slices. This intensity should go all the way up, but instead it's disappearing due to the magnon velocity (slope of band) being higher than the slope of the kinematic constraint boundary:

image

so definitely visualizing this is a challenge! One option is to plot the model beneath the smaller-coverage data like here:

bin_effect_tutorial-15

so the model fills in the gaps (but this is separate from fitting).

bingli621:
You are absolutely correct. If the data set is measured with enough statistics, then 4D fitting should be the way to go, if your data contains little "background" signals. For the magnetism community,  it means low-lying phonons, sample holder, and any sample environment in the beam (CCR, pressure cell, etc.). 
But realistically, most data came with huge error bars (considering that neutron is a weak probe. Also a typical question to ponder on during an experiment is, should we count longer to improve the statistics or measure at another temperature). Binning is the last resort to enhance the signal, reduce the background, and shorten the error bars. The price you pay is losing some information after binning.
The contour plots don't show the error bars, but this information should definitely be used in the fitting. I am not sure if it is always better to fit a noisy 4D dataset, or a less noisy binned dataset. It depends on the statistics.

PR #185, opened 2023-11-06
Dipole LSWT Optimizations

ddahlbom:
A few optimizations for dipole LSWT (starting with removal of allocations associated with biquadratic interactions).


ddahlbom:
The transformed dipole and quadrupole exchange matrices are now precomputed. The speedup for Hamiltonian construction is about 1.35x for dipole, and about 200x for quadrupole (on the test cases I tried).

I made `PairInteraction` mutable to do this in as simple a manner as possible. We could instead just reconstruct the entire `PairInteraction`. Let me know your preferences @kbarros .

ddahlbom:
I realize now that the single ion anisotropy can be precomputed as well (there's no phase).

ddahlbom:
Reverted to non-precalculation of onsite terms and made `PairCoupling` immutable. This should be good to go.

PR #191, opened 2023-11-13
Allow reshaping an inhomogeneous system to a single unit cell

kbarros:
This feature will support the upcoming LSWT-KPM capability developed by @hlane33.

### DIFF
--- a/src/Reshaping.jl
+++ b/src/Reshaping.jl
@@ -34,35 +34,35 @@ function reshape_supercell(sys::System{N}, shape) where N
 end
 
 
-# Transfer homogeneous interactions from `orig` to reshaped `sys`.
+# Transfer interactions from `src` to reshaped `sys`.
 #
-# Frequently `orig` will refer to `sys.origin`, associated with the original
+# Frequently `src` will refer to `sys.origin`, associated with the original
 # crystal, which will make symmetry analysis available. In this case, the
 # process to set a new interaction is to first modify `sys.origin`, and then to
 # call this function.
-function set_interactions_from_origin!(sys::System{N}, orig::System{N}) where N
+function set_interactions_from_origin!(sys::System{N}, src::System{N}) where N
     new_ints = interactions_homog(sys)
 
     for new_i in 1:natoms(sys.crystal)
         # Find `src` interaction either through an atom index or a site index
-        if is_homogeneous(orig)
-            i = map_atom_to_orig_crystal(sys.crystal, new_i, orig.crystal)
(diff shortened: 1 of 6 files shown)


src/Reshaping.jl
```diff
@@ -32,19 +34,35 @@ function reshape_supercell(sys::System{N}, shape) where N
 end
 
 
-# Transfer homogeneous interactions from `sys.origin` to reshaped `sys`
-function set_interactions_from_origin!(sys::System{N}) where N
-    origin = sys.origin
-    new_ints  = interactions_homog(sys)
-    orig_ints = interactions_homog(origin)
+# Transfer homogeneous interactions from `orig` to reshaped `sys`.
```
ddahlbom, line 37: Should "homogenous" be removed from the description here?

src/Reshaping.jl
```diff
@@ -8,6 +8,8 @@ periodicity of a requested supercell. The columns of the ``3×3`` integer matrix
 original crystal lattice vectors.
 """
 function reshape_supercell(sys::System{N}, shape) where N
+    is_homogeneous(sys) || error("Cannot reshape system with inhomogeneous interactions.")
```
ddahlbom, line 11: The point of this is to prevent users from doing this? In other words, you have to know about `reshape_supercell_aux`.

ddahlbom (commented):
This looks like a nice solution to me.

src/Reshaping.jl
kbarros: Yes. I assume we don't want people reshaping inhomogeneous systems in general. This PR makes an exception for LSWT, which is a very specialized use case.

kbarros:
I am thinking this PR is still correct... probably we need to iterate with a real test case.

src/Reshaping.jl
kbarros: Updated

kbarros:
Had a helpful conversation with @ddahlbom. We realized there are some possible simplifications. Now `reshape_supercell_aux` should work more generally. Ready for testing by @hlane33 !

src/Reshaping.jl
kbarros: Update -- there's a bit more flexibility now. For example, with the latest commits, it becomes possible to call `repeat_periodically` on an inhomogeneous system.

kbarros:
Merging now to facilitate @hlane33's work. It needs more testing, but the use cases are somewhat specialized, so we can deal with any errors as they arise.

PR #194, opened 2023-11-21
Update unit tests to avoid warnings

kbarros:
Remove internal functions `*_primitive_crystal`. Instead, it is is recommended to use the conventional unit cell, and later call `reshape_supercell`.


kbarros:
Fixes to `merge_correlations` are now split into separate PR, https://github.com/SunnySuite/Sunny.jl/pull/196.

Issue #149, opened 2023-08-28
Integration assistance


ddahlbom:
Provide:
1) Function to analyze `Langevin` time step
2) Function to analyze `ImplicitMidpoint` time step
3) Function to estimate decorrelation times
4) Functions to output samples and trajectories in newly allocated arrays
5) Tutorial demonstrating use of the above

kbarros:
Regarding 1 and 2, work is in progress here: https://github.com/SunnySuite/Sunny.jl/pull/208

Regarding 3, I wonder if we can leverage one of these packages:

- https://github.com/m3g/BlockAverage.jl
- https://github.com/carstenbauer/BinningAnalysis.jl

kbarros:
Closing this as completed following https://github.com/SunnySuite/Sunny.jl/pull/208. We could still use a tutorial to illustrate [BinningAnalysis](https://github.com/carstenbauer/BinningAnalysis.jl), but that seems independent of immediate Sunny development.

PR #208, opened 2023-12-30
Suggest integration timestep

kbarros:
Collect statistics from numerical integration to provide suggestions about the choice of dt.

### DIFF
--- a/examples/03_LLD_CoRh2O4.jl
+++ b/examples/03_LLD_CoRh2O4.jl
@@ -25,35 +25,44 @@ minimize_energy!(sys)
 sys = resize_supercell(sys, (10, 10, 10))
 @assert energy_per_site(sys) ≈ -2J*S^2
 
-# Use the stochastic Landau-Lifshitz dynamics to thermalize system into
-# equilibrium at finite temperature via a [`Langevin`](@ref) integrator. The
-# timestep ``Δt`` controls integration accuracy. In `:dipole` mode, it should be
-# inversely proportional to the largest effective field in the system. For
-# CoRh₂O₄, this is the antiferromagnetic exchange ``J`` times the spin magnitude
-# ``S=3/2``. The dimensionless parameter ``λ`` determines the magnitude of
-# Langevin noise and damping terms. A reasonable choice is `λ = 0.2`. The
-# temperature `kT` is linked to the magnitude of the Langevin noise term via a
-# fluctuation-dissipation theorem.
+# We will be using a [`Langevin`](@ref) spin dynamics to thermalize the system.
+# This involves a damping term of strength `λ` and a noise term determined by
+# the target temperature `kT`.
 
-Δt = 0.05/abs(J*S)   # Integration timestep
-λ  = 0.2             # Dimensionless damping time-scale
+λ  = 0.2             # Magnitude of damping (dimensionless)
 kT = 16 * meV_per_K  # 16K, a temperature slightly below ordering
+
+# Use [`suggest_timestep`](@ref) to obtain a reasonable integration timestep.
+# The spin configuration in `sys` should ideally be equilibrated for the target
+# temperature `kT`, but the current, energy-minimized configuration will also
+# work reasonably well. Usually `tol=1e-2` is good tolerance to numerical error.
+
+suggest_timestep(sys; tol=1e-2, λ, kT)
+
+# We now have all data needed to construct a Langevin integrator.
+
+Δt = 0.025
 langevin = Langevin(Δt; λ, kT);
 
-# Because the magnetic order has been initialized correctly, relatively few
-# additional Langevin time-steps are required to reach thermal equilibrium.
+# Now run a dynamical trajectory to sample spin configurations. Keep track of
+# the energy per site at each time step.
 
 energies = [energy_per_site(sys)]
 for _ in 1:1000
     step!(sys, langevin)
     push!(energies, energy_per_site(sys))
 end
-plot(energies, color=:blue, figure=(size=(600,300),), axis=(xlabel="Time steps", ylabel="Energy (meV)"))
 
-# After running a Langevin trajectory, it is a good practice to call
-# [`check_timestep`](@ref).
+# Calling [`check_timestep`](@ref) indicates that our choice of `Δt` was a
+# little smaller than necessary; increasing it will improve efficiency.
 
-check_timestep(langevin; tol=1e-2)
+check_timestep(sys, langevin; tol=1e-2)
+langevin.Δt = 0.042
+
+# The energy per site has converged, which suggests that the system has reached
+# thermal equilibrium.
+
+plot(energies, color=:blue, figure=(size=(600,300),), axis=(xlabel="Timesteps", ylabel="Energy (meV)"))
 
 # Thermal fluctuations are apparent in the spin configuration.
 
@@ -63,8 +72,8 @@ plot_spins(sys; color=[s'*S_ref for s in sys.dipoles])
 # ### Instantaneous structure factor
 
 # To visualize the instantaneous (equal-time) structure factor, create an object
-# [`instant_correlations`](@ref) and use [`add_sample!`](@ref) to accumulated
-# data for the equilibrated spin configuration.
+# [`instant_correlations`](@ref). Use [`add_sample!`](@ref) to accumulate data
+# for each equilibrated spin configuration.
 
 sc = instant_correlations(sys)
 add_sample!(sc, sys)    # Accumulate the newly sampled structure factor into `sf`
@@ -101,7 +110,7 @@ formfactors = [FormFactor("Co2")]
 instant_formula = intensity_formula(sc, :perp; formfactors)
 iq = instant_intensities_interpolated(sc, qs, instant_formula);
 
-# Plot the resulting intensity data ``I(𝐪)``. The color scale is clipped to 50%
+# Plot the resulting intensity data ``𝒮(𝐪)``. The color scale is clipped to 50%
 # of the maximum intensity.
 
 heatmap(q1s, q2s, iq;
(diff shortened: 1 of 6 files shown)


kbarros:
Intended to address first part of https://github.com/SunnySuite/Sunny.jl/issues/149.

As currently implemented, this PR collects statistics of `grad E` over the course of a dynamical trajectory and stores them in the `Langevin` struct. An alternative implementation would have `check_timestep` take a `System`, and suggest a timestep based on a single `grad E` calculation. Switching to that alternative implementation would be simpler, and allows the possibility of getting a `dt` estimate prior to running dynamics. A disadvantage to switching is that the estimates might become worse for small systems where there are less statistics (?).

Another question is how to define an interpretable error tolerance parameter, `tol`. Since Heun is weak first-order accurate, the error in statistical observables likely decays as `1 / dt`. It might be the case that if we switch to implicit-midpoint as our Langevin integrator, the error decays faster, as `1 / dt^2`. We could determine this with some numerical experiments.

Prior to merging, we should also add support for `ImplicitMidpoint`.

src/Integrators.jl
```diff
@@ -1,53 +1,178 @@
...
+    #
+    # The total error (err1 + err2) should be less than the target tolerance.
+    # After some algebra, this implies,
+    #
+    # dt ≲ sqrt(tol / (c₁² |drift|² + c₂² λ² kT²))
+    #
+    # for some empirical constants c₁ and c₂.
+
+    c1 = 1.0
+    c2 = 1.0
```
ddahlbom, line 149: Is it correct to assume that the appropriate constants are model-dependent?

examples/03_LLD_CoRh2O4.jl
```diff
@@ -26,13 +26,17 @@ sys = resize_supercell(sys, (10, 10, 10))
...
+# timestep ``Δt`` controls integration accuracy. In `:dipole` mode, it should be
+# inversely proportional to the largest effective field in the system. For
+# CoRh₂O₄, this is the antiferromagnetic exchange ``J`` times the spin magnitude
+# ``S=3/2``. The dimensionless parameter ``λ`` determines the magnitude of
+# Langevin noise and damping terms. A reasonable choice is `λ = 0.2`. The
+# temperature `kT` is linked to the magnitude of the Langevin noise term via a
+# fluctuation-dissipation theorem.
+
+Δt = 0.05/abs(J*S)   # Integration timestep
+λ  = 0.2             # Dimensionless damping time-scale
```
ddahlbom, line 39: What was the motivation for taking λ to 0.2? More reasonable decorrelation times? (I think I used 0.2 for FeI2.)

examples/03_LLD_CoRh2O4.jl
```diff
@@ -46,6 +50,11 @@ for _ in 1:1000
 end
 plot(energies, color=:blue, figure=(size=(600,300),), axis=(xlabel="Time steps", ylabel="Energy (meV)"))
 
+# After running a Langevin trajectory, it is a good practice to call
```
ddahlbom, line 53: Asked about this on slack. It seems reasonable that a trajectory should be collected for statistics. I guess there are some question about whether this implicit definition of a trajectory is the best interface. Will comment more outside of code.

src/Integrators.jl
```diff
@@ -83,7 +208,19 @@ function step!(sys::System{0}, integrator::Langevin)
...
+    # Collect statistics about the magnitude of the drift term squared. For the
+    # energy-conserving part, it is important to use |∇E|² rather than |s×∇E|²,
+    # because the former gives direct information about the frequency of
+    # oscillation, while the latter may artificially vanish (consider, e.g., a
+    # spin nearly aligned with an external field).
+    for i in eachsite(sys)
+        s0 = sys.κs[i] # == norm(s[i])
+        ∇E² = ∇E[i]' * ∇E[i]
+        accum!(integrator.drift_squared, (1 + (s0*λ)^2) * ∇E²)
+    end
```
ddahlbom, line 221: Have you benched the effect of this? Perhaps it should be optional? (I imagine the cost is much less than the calculation of the gradient, but I'm curious about the rough order of magnitude.)

ddahlbom (commented):
This looks like it will be very useful.

Perhaps the collection of statistics should be something that a user consciously chooses to do. I assume also there's some small performance hit associated with calculating the errors.

ddahlbom:
I think that collecting statistics for a chosen time steps is reasonable, even if in an ideal world a user would like to have a function that takes one step and gets a suggestion. I assume the statistics will vary a lot depending on whether the system is equilibrated or not, and the user should probably be encouraged to do a small study for the situation that they are interested in studying. I think the report that this provides is very useful.

src/Integrators.jl
kbarros: I think so. We are trying to find a mapping between error and dt, but this depends on the model and the quantity used to define "error". Since it's fundamentally ambiguous, I defaulted to a simple choice here, which seemed to work well for our examples.

The ratio between c2/c1 particularly hard to select. It seems natural that, when T gets "big enough", we want dt to scale like T at a fixed tolerance. But how should we define "big enough" relative to other energy scales in the system?

examples/03_LLD_CoRh2O4.jl
kbarros: It is problem dependent, but I suspect that larger λ speeds decorrelation in the sense that "gradient descent" becomes more efficient. But in other ways it could slow things, because you're damping out spin waves that may carry information ballistically.

kbarros:
> I assume the statistics will vary a lot depending on whether the system is equilibrated or not, and the user should probably be encouraged to do a small study for the situation that they are interested in studying. I think the report that this provides is very useful.

A worst case scenario is maybe a single-site problem, where `norm(HZ)` could be any eigenvalue of the single-ion Hamiltonian depending on configuration.

Probably for a system with multiple sites, a random spin configuration will give a good sense of the typical dt required.

Conversely, for a local Monte-Carlo sampler, I don't see anyway around the need to collect statistics over a trajectory. Given that this type of interface must exist, it might make sense to use it consistently for Langevin sampling.

kbarros:
I tested the alternate implementation, which suggests `Δt` based on the energy gradient of a single snapshot. 

Results for the dynamics in our two tutorials. The suggested Δt below varies according to the type of spin snapshots used in the estimation:

**Tutorial03, CoRh2O4**

Δt ~ 0.02528 (Polarized in z)
Δt ~ 0.04959 (T=Inf)
Δt ~ **0.02528** (T=0)
Δt ~ **0.03933** (T=16K equilibrium snapshot)
Δt ~ 0.04201 (T=16K, equilibrium averaged)

**Tutorial04, FeI2**

Δt ~ 0.1482 (Polarized in z)
Δt ~ 0.05107 (T=Inf)
Δt ~ **0.02734** (T=0)
Δt ~ **0.03794** (T=3.5meV, equilibrium snapshot)
Δt ~ 0.03885 (T=3.5meV, equilibrium averaged)

Key points:

1. Energy gradient is larger for configurations at lower temperatures, leading to smaller suggested dt.
2. At finite temperatures, one should in principle average over many equilibrium snapshots. For these particular systems, with 64 sites, this time averaging isn't overkill, because a single snapshot already includes an average over sites.
3. The suggested dt at T=0 is approximately 30% lower than at temperatures around than transition.

I think a reasonable advice to users is "first run `minimize_energy!` and then call `suggest_timestep`. This is a simple interface, and can be done prior to any time-integration.

kbarros:
@ddahlbom  pointed out that there might be a different trend with frustrated systems. This can be verified in the SpinW tutorial:

SW08, Kagome antiferro:

Δt ~ 0.025 (Polarized in z)
Δt ~ 0.045 ± 0.01 (T=Inf)
Δt ~ 0.05 (T=0)

src/OnlineStatistics.jl
```diff
@@ -0,0 +1,27 @@
+
+mutable struct OnlineStatistics{T}
```
ddahlbom, line 2: Do you plan to keep this around?

src/Integrators.jl
```diff
@@ -85,6 +91,127 @@ mutable struct ImplicitMidpoint
...
+This implies that the `tol` argument may actually scale like the _square_ of the
+true numerical error, and should be selected with this in mind.
+"""
+function suggest_timestep(sys::System{N}, integrator::Langevin; tol) where N
+    (; Δt, λ, kT) = integrator
+    suggest_timestep_aux(sys; tol, Δt, λ, kT)
+end
+function suggest_timestep(sys::System{N}, integrator::ImplicitMidpoint; tol) where N
+    (; Δt) = integrator
+    suggest_timestep_aux(sys; tol, Δt, λ=0, kT=0)
```
ddahlbom, line 131: Do you want to put a factor of 2 in here? That has been the rule of thumb I've generally used for setting the ImplicitMidpoint time step.

examples/04_GSD_FeI2.jl
```diff
@@ -131,6 +143,11 @@ for _ in 1:10_000
     step!(sys_large, langevin)
 end
 
+# With this increase in temperature, the suggested timestep has increased slightly.
+
+suggest_timestep(sys_large, langevin; tol=1e-2)
+langevin.Δt = 0.040;
```
ddahlbom, line 149: I think this interface is good. I guess the pattern is new in Sunny: allowing a field to be left unset in a struct and requiring that it be set before use. But I think I prefer this to, say, having a setter function.

ddahlbom (commented):
I think this looks good. It makes it very easy for users. 

I wonder if a factor of two or so should be applied to the ImplicitMidpoint recommendation.

Relatedly, it seems like it would be useful to study the number of steps to converge when using ImplicitMidpoint as another reference point. I believe there is a point when the number of steps it takes to converge starts growing quite rapidly, and a good choice is generally to pick a time step as large as possible before that growth kicks in. But that is a more complicated analysis. Whatever is recommended in this version would certainly be safe, I think.

examples/04_GSD_FeI2.jl
```diff
@@ -141,18 +158,21 @@ end
...
-# an integration step size, a target number of ωs to retain, and a maximum
-# energy ω to resolve. For the time step, twice the value used for the Langevin
-# integrator is usually a good choice.
+# The function [`dynamical_correlations`](@ref) creates an object to store
+# sampled correlations. The integration timestep `Δt` used for measuring
+# dynamical correlations can be somewhat larger than that used by the Langevin
+# dynamics. We must also specify `nω` and `ωmax`, which determine the
+# frequencies over which intensity data will be collected.
 
-sc = dynamical_correlations(sys_large; Δt=2Δt, nω=120, ωmax=7.5)
```
ddahlbom, line 150: I see the two is left here.

From an interface standpoint, I guess it's a bit confusing, because in the typical use case the user never sets a time step for ImplicitMidpoint. But they have the option of estimating if they so choose, and that estimate will be half of what is used here.

src/OnlineStatistics.jl
kbarros: I think it could later be useful in other places, but I removed the `include` from Sunny.jl.

PR #217, opened 2024-01-09
Asymmetric correlations (continued from #174 )

Lazersmoke:
This is #174 but fixed up and rebased onto modern Sunny


src/Integrators.jl
```diff
@@ -266,6 +266,7 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
 
     s = sys.dipoles
     (; Δt, atol) = integrator
+    isnan(Δt) && return
```
kbarros, line 269: This check is already made in `check_timestep_available()` above.

Lazersmoke:
Here is a list of what this PR adds/changes (can be condensed into some `versions.md` lines):

1. Adds `negative_energies=true` support to the binning interface. (Required to be able to evaluate the sum rule, which requires integrating over `[-fs/2,fs/2]`, and make [this test](https://github.com/SunnySuite/Sunny.jl/pull/217/files#diff-2fd9a341447572a46615b37ad827bdf63231949c8cae721bb5adf99880538fd3R1) pass)
2. Corrects the overall scale of intensities returned by `intensities_binned` to correctly incorporate the bin size. (Required for binned sum rule to be satisfied)
3. Adds concept of "`unilateral_to_bilateral` transformation" to the `Contraction`s interface. (Required for agreement between LSWT (`=true`) and SampledCorrelations (`=false`). See [here](https://github.com/SunnySuite/Sunny.jl/pull/217/files#diff-903cd4af93317fd589d236c23695740eef7f0a7de0768c3a7a9415a25683f4caR189-R194) for the important part. Amounts to a factor of x2 in intensities in some cases)
4. [Documents](https://github.com/SunnySuite/Sunny.jl/pull/217/files#diff-903cd4af93317fd589d236c23695740eef7f0a7de0768c3a7a9415a25683f4caR102)/corrects the internal handling of indices of correlations in the contractions interface to be self-consistent.
5. Adds support function `intensities_band_structure`, which returns the LSWT intensities in a neat struct `bs::BandStructure` with `bs.dispersion` and `bs.intensities` fields. The idea is that the normalization of these "delta function" intensities requires some [verbal explanation (currently located in a comment)](https://github.com/SunnySuite/Sunny.jl/pull/217/files#diff-31d12bb3098b7a623c4814c24ff8509e24693935e9d859f8f80000e026761e11R362-R366), which can be given in a docstring for `BandStructure` later. (also adds show instance for this previously internal type).
6. Allows querying both (Sx,Sy) and (Sy,Sx) correlations, which are different in general. (Although in certain special cases, such as classical dynamics they coincide up to time reversal/some other symmetry). There is a [well-documented test](https://github.com/SunnySuite/Sunny.jl/pull/217/files#diff-2fd9a341447572a46615b37ad827bdf63231949c8cae721bb5adf99880538fd3R18) which compares this to a reference time-domain algorithm.
7. Fixes up the classical intensities calculation by: a. fixing wraparound bug using zero padding, b. inserting a `1/natoms` factor, c. replacing an extraneous `1/n_omega` with the correct `1/(T - |t|)`, d. enabling an optional cosine-squared window function which may be useful for systems which do not naturally decorrelate. These changes (correctly) alter the normalization of the intensities, and this change is [documented in the test suite](https://github.com/SunnySuite/Sunny.jl/pull/217/files#diff-05f6b3d8fbb2bce4da7426595baf85c2707984667c266a15eb399c51115bde14R100-R103). (Required to make the test from (6.) pass)
8. Optimizes memory allocation when using LSWT-with-kernel
9. Adds a [well-documented test](https://github.com/SunnySuite/Sunny.jl/pull/217/files#diff-05f6b3d8fbb2bce4da7426595baf85c2707984667c266a15eb399c51115bde14R26) of the "polyatomic sum rule," which is more stringent than existing sum rule tests. The existing "monoatomic" tests still exist and are now [better documented](https://github.com/SunnySuite/Sunny.jl/pull/217/files#diff-2fd9a341447572a46615b37ad827bdf63231949c8cae721bb5adf99880538fd3R116)

Lazersmoke:
#237 includes points (1) (2) (7b) (7c, sort of) and (9) from the above list

kbarros:
This feature is currently out of scope, but can be revisited if there is sufficient user demand.

PR #228, opened 2024-01-26
Add checks on Langevin parameters

kbarros:
Small damping can lead the large statistical error, and the user should be guided towards ImplicitMidpoint for now.


ddahlbom (approved):
Looks sensible to me.

PR #229, opened 2024-01-27
LSWT corrections to the classical energy and ordered moment 

Hao-Phys:
We have integrated LSWT corrections into the classical energy and ordered moment for both :SUN and :dipole modes. The numerical integrations are performed using HCubature.jl.

However, we have encountered an issue. When attempting to report the classical energy during correction calculations, the energy_per_site(swt.sys) function yields nonsensical results. This discrepancy is likely attributed to the rotation of all interactions. Shall we explore potential solutions to address this problem?


kbarros:
Yes, the plan is to undo the rotations in the `swt.sys` and store them in a new `set.interactions` field. I can do it today.

PR #231, opened 2024-02-02
Support inhomogeneous observables

Lazersmoke:
Waiting on tests to run; it may fail due to things being incorrect before!

### DIFF
--- a/src/Operators/Observables.jl
+++ b/src/Operators/Observables.jl
@@ -1,7 +1,15 @@
 
+# An observable is either an operator, or a field of operators.
+# If it's a field of operators, then it's defined with reference
+# to some System that users of this type are responsible for keeping track of.
+ObservableOperator = Union{LinearMap, Array{LinearMap,4}}
+
+observable_at_site(op::LinearMap,site) = op
+observable_at_site(op::Array{LinearMap,4},site) = op[site]
+
 struct ObservableInfo
     # Correlation info (αβ indices of S^{αβ}(q,ω))
-    observables    :: Vector{LinearMap}  # Operators corresponding to observables
+    observables    :: Vector{ObservableOperator}  # Operators corresponding to observables
     observable_ixs :: Dict{Symbol,Int64} # User-defined observable names
     correlations   :: SortedDict{CartesianIndex{2}, Int64}  # (α, β) to save from S^{αβ}(q, ω)
 end
@@ -25,6 +33,27 @@ function Base.show(io::IO, ::MIME"text/plain", obs::ObservableInfo)
         println(io)
     end
     printstyled(io,"")
+end
+
+function multiply_by_g_factor!(sys,obs)
+    g = sys.gs
+
+    # Easy/common case
+    if allequal(g)
+        for k = 1:num_observables(obs)
+            obs.observables[k] = obs.observables[k] * g[1]
+        end
+        return
+    end
+
+    for k = 1:num_observables(obs)
+        gA = Array{LinearMap,4}(undef,size(g))
+        for site = eachsite(sys)
+            A0 = observable_at_site(obs.observables[k],site)
+            gA[site] = A0 * g[site]
+        end
+        obs.observables[k] = gA
+    end
 end
 
 # TODO: Add/unify docs about allowed list of observables. See comment here:
--- a/src/SampledCorrelations/CorrelationSampling.jl
+++ b/src/SampledCorrelations/CorrelationSampling.jl
@@ -1,22 +1,19 @@
-function observable_values!(buf, sys::System{N}, ops; apply_g = true) where N
+function observable_values!(buf, sys::System{N}, ops) where N
     if N == 0
-        for site in eachsite(sys), (i, op) in enumerate(ops)
-            dipole = sys.dipoles[site]
-            if apply_g
-              dipole = sys.gs[site] * dipole
+        for (i, op) in enumerate(ops)
+            for site in eachsite(sys)
+                A = observable_at_site(op,site)
+                dipole = sys.dipoles[site]
+                buf[i,site] = A * dipole
             end
-            buf[i,site] = op * dipole
         end
     else
         Zs = sys.coherents
-        #num_ops =  size(ops′, 3)
-        #ops = reinterpret(SMatrix{N, N, ComplexF64, N*N}, reshape(ops′, N*N, num_ops))
-
-        # SQTODO: This allocates :(
         for (i, op) in enumerate(ops)
-          matrix_operator = convert(Matrix{ComplexF64},op)
             for site in eachsite(sys)
-                buf[i,site] = dot(Zs[site], matrix_operator, Zs[site])
+                A = observable_at_site(op,site)
+                #matrix_operator = convert(Matrix{ComplexF64},A)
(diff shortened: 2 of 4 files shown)


kbarros:
Thanks a lot for putting this together. Should we also switch to `g=2` for some of the tests in this PR to exercise the new functionality?

Lazersmoke:
Squishing the last of the bugs now, and for this PR and the other one which updates the intensities scale, I'm doing this sequence:

1. Keep the old test data unchanged (for now), and do `@test new / g_factor^2 == old` to make it pass and be correct
2. Merge, so that line gets in the commit history
3. Update the test data in a later commit, and then we'll be back to `@test new == old`

which I think is reasonable.

src/SampledCorrelations/CorrelationSampling.jl
```diff
@@ -1,22 +1,19 @@
-function observable_values!(buf, sys::System{N}, ops; apply_g = true) where N
+function observable_values!(buf, sys::System{N}, ops) where N
     if N == 0
-        for site in eachsite(sys), (i, op) in enumerate(ops)
-            dipole = sys.dipoles[site]
-            if apply_g
-              dipole = sys.gs[site] * dipole
```
kbarros, line 6: Moving the `apply_g` check out of this branch is a big win.

src/SampledCorrelations/CorrelationSampling.jl
```diff
@@ -1,22 +1,19 @@
...
+        for (i, op) in enumerate(ops)
+            for site in eachsite(sys)
+                A = observable_at_site(op,site)
+                dipole = sys.dipoles[site]
+                buf[i,site] = A * dipole
```
kbarros, line 7: Just an FYI: More generally, we will eventually want _nonlinear_ maps of dipole for observables. This could naturally be achieved with Stevens polynomials. But that discussion can happen later.

kbarros:
Looks good to me. @ddahlbom would you mind also taking a look?

I guess we're waiting on https://github.com/SunnySuite/Sunny.jl/issues/232 before adding a test with a general 3x3 g-tensor?

src/SampledCorrelations/CorrelationSampling.jl
Lazersmoke: Ah yes I see. Observables are simply different for dipole mode vs SU(N) it seems. (Mainly because dipole mode e.g. doesn't include quadrupoles when S=1)

Lazersmoke:
#232 only impacts 1/4 modes so we can add three tests and a broken test now

src/SampledCorrelations/CorrelationSampling.jl
```diff
@@ -1,22 +1,19 @@
...
-        #num_ops =  size(ops′, 3)
-        #ops = reinterpret(SMatrix{N, N, ComplexF64, N*N}, reshape(ops′, N*N, num_ops))
-
-        # SQTODO: This allocates :(
         for (i, op) in enumerate(ops)
-          matrix_operator = convert(Matrix{ComplexF64},op)
             for site in eachsite(sys)
-                buf[i,site] = dot(Zs[site], matrix_operator, Zs[site])
+                A = observable_at_site(op,site)
+                #matrix_operator = convert(Matrix{ComplexF64},A)
```
ddahlbom, line 15: Can this line be deleted?

ddahlbom (commented):
I agree this looks good. It's nice to completely move the g-factor into the observable construction/processing.

kbarros:
> #232 only impacts 1/4 modes so we can add three tests and a broken test now

Adding this sounds great. Could we additionally have Sunny throw an informative error if it detects the unsupported case? "Anisotropic g-tensor is not currently supported when [...]".

Lazersmoke:
Now closes #232

test/test_operators.jl
```diff
@@ -345,3 +345,48 @@ end
...
+        # Dipole magnetization observables (classical)
+        sc = instant_correlations(sys)
+        add_sample!(sc,sys)
+
+        formula = intensity_formula(sc,:full)
+        corr_mat = instant_intensities_interpolated(sc,[[0,0,0]],formula,interpolation = :round)[1]
+
+        # Compute magnetization correlation "by hand", averaging over sites
+        mag_corr = sum([sys.gs[i] * sys.dipoles[i] * (sys.gs[j] * sys.dipoles[j])' for i = 1:2, j = 1:2]) / Sunny.natoms(cryst)
+        @test_broken isapprox(corr_mat,mag_corr) # Sunny bug: classical intensity off by natoms factor
```
kbarros, line 371: This @test_broken is a nice trick to force us to revisit later.

src/SpinWaveTheory/SpinWaveTheory.jl
```diff
@@ -205,6 +209,12 @@ function swt_data_dipole!(sys::System{0})
         # Precalculate operators for rotating Stevens coefficients.
         V = operator_for_stevens_rotation(2, R)
 
+        # Observables are 1x3 row vectors
+        for μ = 1:num_observables(obs)
+            row = observable_at_site(obs.observables[μ],CartesianIndex(1,1,1,atom))
+            observables_localized[:,:,μ,atom] .= Matrix(row) * R
```
ddahlbom, line 215: If `row` is a row vector, how is Matrix(row) * R a 2-dimensional object? (I think I get the big idea, but I'm a bit confused by the implementation.)

ddahlbom (commented):
Thanks for putting this together! The approach looks good to me. I'm a bit confused by some implementation details when it comes to getting the g-factor into dipole SWT. (I believe the implementation is correct. I just want to make sure I follow the details.)

src/Operators/Observables.jl
```diff
@@ -27,22 +35,43 @@ function Base.show(io::IO, ::MIME"text/plain", obs::ObservableInfo)
     printstyled(io,"")
 end
 
+function magnetization_observables(g,N)
+    dipoles(single_g,N) = if N == 0
+        dipole_component(i) = FunctionMap{Float64}(s -> (single_g*s)[i],1,3)
```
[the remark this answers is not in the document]
Lazersmoke: Yes, `dx = dipole_component(1)` will be a 1x3 linear map, which you can just use as a matrix. When it acts, `dx * S` on a vector `S`, it will do `(s -> (single_g*s)[1])(S)` which becomes `(single_g * S)[1]`. So it's a lot like just taking `single_g[1,:]`, and maybe we could try that instead and compare performance. The reason to do it as a LinearMap is that you might be able to implement this faster than a matrix multiply! For example if `single_g` is `I(3)`, it may be optimized to be faster

src/SpinWaveTheory/SpinWaveTheory.jl
Lazersmoke: It's a `1x3` matrix, so it's teeeeeechnically 2-dimensional. The `size(observables_localized,1)` is `1`

ddahlbom: Gotcha. Thanks.

kbarros:
This will resolve both https://github.com/SunnySuite/Sunny.jl/issues/236 and https://github.com/SunnySuite/Sunny.jl/issues/232.

PR #234, opened 2024-02-06
Catch errors during show(System)

Lazersmoke:
Now that `energy(sys)` sometimes errors (e.g. if dipoles are not normalized), typing in `sys` at the REPL can clobber the terminal with a huge error message. This PR makes it so the error message is relatively tame

### DIFF
--- a/src/System/System.jl
+++ b/src/System/System.jl
@@ -97,10 +97,10 @@ function lattice_to_str(sys::System)
 end
 
 function energy_to_str(sys::System)
-    try
+    if is_valid_normalization(sys)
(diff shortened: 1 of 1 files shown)


kbarros:
I modified the PR slightly so that it has the following behavior:
```julia
sys = System(cryst, dims, [SpinInfo(1; S, g=1)], :dipole)
display(sys)
#=
System [Dipole mode]
Lattice (1×1×1)×1
Energy per site 0
=#

sys.dipoles[1,1,1,1] *= 2
display(sys)
#=
System [Dipole mode]
Lattice (1×1×1)×1
[Incorrectly normalized spin state!]
=#
```

PR #239, opened 2024-02-19
Thermostat for implicit midpoint

ddahlbom:
This PR adds a thermostat to the implicit midpoint integrator. The interface changes are minimal: `ImplicitMidpoint` now optionally takes `kT` and `λ`. By default both these keywords are set to zero.

All the finite temperature sampling tests (against analytical solutions for anisotropies, spin chains, etc.) now also exercise the implicit midpoint integrator.

We may in the future wish to redo some naming. (Having `Langevin` specify a Heun integrator makes less sense now.) But these changes should not break anyone's code and they will hopefully be useful to @Lazersmoke's efforts.


ddahlbom:
Will update docstrings soon.

kbarros:
Looks very nice to me. The only question I have is about whether this slows down some workflows -- ImplicitMidpoint timesteps will be the bottleneck for many calculations. If so, there is an optimization idea in comment above.

src/Integrators.jl
```diff
@@ -319,6 +326,15 @@ end
     a - Z * ((Z' * a) / (Z' * Z))
 end
 
+function rhs!(ΔZ, Z, ξ, HZ, integrator, sys::System{N}) where N
+    (; kT, λ, Δt) = integrator
+    for site in eachsite(sys)
+        ΔZ′ = -im*√(2*Δt*kT*λ)*ξ[site] - Δt*(im+λ)*HZ[site]
```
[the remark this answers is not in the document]
ddahlbom: Good point. Thanks. Will change.

ddahlbom:
Yes, I was concerned about that too. I'll make the change you suggested and do some simple benchmarking. I guess if it is noticeably slower, we can go back to dispatching on the type level (so instead of relying on branch prediction we just do one function lookup). Maybe have an `ImplicitMidpointLangevin` integrator.

src/Integrators.jl
```diff
@@ -282,7 +289,7 @@ function step!(sys::System{0}, integrator::ImplicitMidpoint)
         # improved midpoint estimator s̄′.
         @. ŝ = normalize_dipole(s̄, sys.κs)
         set_energy_grad_dipoles!(∇E, ŝ, sys)
-        @. s̄′ = s + 0.5 * Δt * rhs_dipole(ŝ, -∇E)
+        @. s̄′ = s + 0.5 * (Δt * rhs_dipole(ŝ, -∇E, λ) + √Δt * rhs_dipole(ŝ, ξ))
```
kbarros, line 292: New call to `rhs_dipole(ŝ, ξ)` might incur nontrivial computational cost. This could be broken into a separate block:

```julia
if kT > 0
    @. s̄′ += √(Δt*2λ*kT) * rhs_dipole(ŝ, ξ))
end
```

And then remove the line `ξ .*= √(2λ*kT)` above.  Note that this scaling of `ξ` as a pure Gaussian is more consistent with the SU(N) case.

ddahlbom:
I took a simple system (S=1, one exchange, onsite anisotropy), thermalized it, and benched two versions of the step function: (1) The original code; and (3) the new code, with suggested optimizations and `λ` and `kT` set to zero.

For dipole:
```
Reference           -- 46.667 μs
New Code            -- 48.666 μs
```

For SU(N)
```
Reference           -- 93.667 μs
New Code            -- 107.292 μs
```

We see a slight slowdown in dipole mode, and serious slowdown in SU(N) mode. Perhaps this makes it worthwhile to keep around the old `ImplicitMidpoint` (with no thermostat) and add an `ImplicitMidpointLangevin`.

Sound reasonable?

ddahlbom:
It turns out the main reason for the performance regression was the fact that the new approach added a call to `proj` in all cases, even when lambda and kT were zero.

The current approach includes a few more branching conditions. The big branching condition is based on whether lambda is nonzero.

These conditionals are kind of ugly, but the result is: no new type, and no performance regression.

kbarros:
I added some commits that fully unify the dipole and SU(N) integration schemes (both Heun and implicit midpoint). This implementation also avoids code redundancy.

Because the convergence check is slightly modified, it is necessary to add new reference data for "Sampled correlations reference".

According to the benchmark below, performance is preserved for implicit midpoint without Langevin thermostat.

```julia

using Sunny, BenchmarkTools

latvecs = lattice_vectors(1, 10, 10, 90, 90, 90)

# Modify for :dipole or :SUN
sys = System(cryst, (1000, 1, 1), [SpinInfo(1, g=2, S=1)], :SUN)
set_exchange!(sys, -1, Bond(1, 1, (1, 0, 0)))
randomize_spins!(sys)

integrator = ImplicitMidpoint(0.02; atol=1e-5)
# suggest_timestep(sys, integrator; tol=1e-3)

function f(sys, integrator)
    randomize_spins!(sys)
    for _ in 1:100
        step!(sys, integrator)
    end
end

integrator.λ = 0
integrator.kT = 0
@btime f(sys, integrator)
# :SUN with S=1
# (new) 14.4 ms <- 14.58 ms (old)
# :dipole
# (new) 6.27 ms <- 6.30 ms (old)

integrator.λ = 0.0001
integrator.kT = 0
@btime f(sys, integrator)
# :SUN with S=1 :: 16.95 ms
# :dipole       :: 6.825 ms

integrator.λ = 0.0001
integrator.kT = 0.0001
@btime f(sys, integrator)
# :SUN with S=1 :: 18.7 ms
# :dipole       :: 7.62 ms
```

src/Integrators.jl
```diff
@@ -222,8 +239,25 @@ end
...
+    else
+        @. Δs = -s × (- Δt*∇E + ξ - Δt*λ*(s × ∇E))
+    end
+end
+
+function rhs_sun!(ΔZ, Z, ξ, HZ, integrator)
+    (; kT, λ, Δt) = integrator
+
+    @. ΔZ = - Δt*(im+λ)*HZ + ξ
+    if !iszero(λ) || !iszero(kT)
```
ddahlbom, line 256: Is the part after the `||` necessary? We should never have the case that `λ=0` but `kT>0`.

ddahlbom (commented):
The reorg looks very nice to me. 

I guess Github doesn't let you commend on unchanged code, but these lines should probably be changed. I don't think there's any reason to have different versions now.
```
function suggest_timestep(sys::System{N}, integrator::Langevin; tol) where N
    (; Δt, λ, kT) = integrator
    suggest_timestep_aux(sys; tol, Δt, λ, kT)
end
function suggest_timestep(sys::System{N}, integrator::ImplicitMidpoint; tol) where N
    (; Δt) = integrator
    suggest_timestep_aux(sys; tol, Δt, λ=0, kT=0)
end
```

src/Integrators.jl
Lazersmoke: I could imagine having that case! It represents the weak-coupling limit :) so for example if I'm running ImplicitMidpoint at finite temperature, but then want to compare to the uncoupled case (which is what Sunny does by default), I would probably just set lambda = 0 (still with finite kT) to achieve this

ddahlbom: I see. Just to be clear, I'm not here saying we disallow users from setting a finite `kT` with `λ=0`. My point is just that the fast path should be taken whenever `λ=0` (since the kT will not practically contribute in the impelmentation). This will be the case if the OR is removed.

ddahlbom:
Looks great to me.

Issue #235, opened 2024-02-07
Unify normalization conventions of dynamical correlation


kbarros:
Dynamical correlations calculated with spin wave theory and with classical dynamics currently follow different normalization conventions with respect to `natoms` (the number of atoms in the magnetic cell). The current state is:

1. SpinW does **not** normalize by `natoms`
2. Our current LSWT code **does** normalize by `natoms`
3. Our current classical dynamics intensity code does **not** normalize by `natoms`.

@Lazersmoke argued on Slack that (2) is correct (even though it differs from the SpinW convention), and that (3) should be changed to match (2).

Here is my lightly edited summary of @Lazersmoke's argument:

>  Ok, so that everyone is on the same page, if I have a quantity S(x,t) with "units" S^2 and I Fourier transform it, then the result will have units S^2/BZ/fs where "per BZ" is [Length]^3 and "per fs" is [Time], and they indicate that the S(q,w) is a density in reciprocal space. In order for the typical density S(q,w) to be on the same order of magnitude as S(x,t), you need to do a fast Fourier transform (which is a sum of N terms with no normalization factor) and then divide by N.
>
> For exactly the same reason, when you add Nm terms together in a spin wave code, you have to divide by Nm to get your intensity to be in the standard S^2/BZ/fs units; you're averaging Nm estimates of the correlation together.
>
> For (once again) the same reason, you need to divide by prod(sys.latsize) when computing spatial correlations.
>
> SpinW is adding together Nm terms and not dividing by Nm [bad]
>
> LSWT is adding together Nm terms and using sqrt_Nm_inv to divide by Nm [good]
>
> LL on main is correctly dividing by prod(sys.latsize) [good], but not by Nm [bad], and for energy axis it's dividing by nomega too many times

Lazersmoke:
This all looks good 👍 and the classical dynamics as implemented in #217 fixes the `Nm` and `n_omega` factors.

One caveat: everything mentioned above includes only up to the output of `calc_intensity`. There are additional normalization considerations that need to be made on case-by-case basis for each of the following functions, which consume the output of `calc_intensity`:
- `intensities_interpolated` (no change)
- `intensities_binned` (normalization having to do with bin width; implemented on 217 )
- `intensities_bands` (normalization of delta functions needs to be documented)
- `intensities_band_structure` (new in 217 ; normalization of delta functions within the `BandStructure` struct needs to be documented)
[...]

PR #240, opened 2024-02-19
Structure factor docs

kbarros:


Lazersmoke:
Hi Kip, thanks for writing this up! This is well on the way to getting it documented what we really compute 👍 

A few comments which I think should be edited in to the document. I've tried to be exhaustive but I may have more comments later. Overall looks really great though!

1. In the expression $A(r_0,0)B(r_0-r,t)$ inside the definition of $C(r,t)$, the time lag $t$ and position lag $r$ are going in opposite directions! In classical Sunny, we [do both temporal and spatial correlations in the same way](https://github.com/SunnySuite/Sunny.jl/blob/35fb5af5ecdd6afa06ecbee2aec4327030b860cb/src/SampledCorrelations/CorrelationSampling.jl#L102), so this should be reflected in the definition in the docs. What we actually use currently is $S^{\alpha\beta}\sim\alpha(r_0 + r,t_0 + t)\beta(r_0,t_0)$.
2. Also in that initial definition of $C(r,t)$, the time dynamics are left unspecified. It's only later that we specify either Heisenberg time evolution, or a classical dynamics of some kind. It would be best to explicitly say that it's undetermined at that early stage of the definition ("where the dependence on time $t$ is to be specified later"), or else the reader may be misled that the classical dynamics uses Heisenberg evolution or vice versa, or that there is a single unique choice of time dynamics for this. This point is extra important for the project I'm working on now, where the $C(r,t)$ is distinct from what's written here; so it would be very useful if this was documented as a point of reference for possible modification.
3. The Lehmann representation is only for spin wave theory (does not apply to the classical case) so it should go on maybe even a separate page (not in `structure-factor.md`) to talk about the SWT calculation. There is also omitted so far the sum $\sum_{\mu,k}$, which is a subtle SWT thing that is specially adapted to the harmonic oscillator. In particular, there is no $\beta$ anywhere in Sunny (except for c2q factor), and until there is, we shouldn't document that there is! The next section about spin-spin correlation covers the $\omega$ part sufficiently.
4. It's best to let the intensive version of the correlations be in "moles of sites" rather than "moles of unit cells", since the latter depends on the choice of crystallographic unit cell. As things are written now, using a 2x larger unit cell makes the signal 2x as intense. With the "moles of sites" definition, all instances of "N_cell" (an unphysical quantity) go away and are replaced with "N_spins" = (total number of magnetic sites in the model), which is a physical quantity.
5. In the cross-section, the `kf/ki` is on the right; we should probably comment that this is moved to the left hand side in practice, since Sunny doesn't do anything with this (ask xiaojian for more details)

A few notes (for the record) about our conventions:
- The $B^\dagger(k) = B(-k)$ (for real $B$) is a convention--and it is the convention that we adopt.
- The expression $C(k,t) = \langle \alpha(k,0) \beta^\dagger(k,t)\rangle$ is compatible with the actually-implemented $S^{\alpha\beta}\sim\alpha(r_0 + r,t_0 + t)\beta(r_0,t_0)$, but not with the listed expression for $C(r,t)$ (from point 1 above)

Lazersmoke:
About the section on the `<MM>` vs `<SS>`. They should coincide when g=1, right? (not g = 2)

I think some of the MM and SS labels are misplaced in the equations.

kbarros:
Addressing the remaining comments from above:

3. This Structure Factor document is a specification for the intended Sunny behavior in calculating quantum observables. We will eventually want a new section describing how to calculate classical correlations, but that is not yet part of mainstream Sunny. The option to include $\beta$ dependence should be applied uniformly across mode (LSWT or `dynamical_correlations`). Apart from bugs in the current implementation, this PR should clearly specify the _intended_ Sunny behavior.
4. Rationale for this normalization convention is given here: https://github.com/SunnySuite/Sunny.jl/issues/235#issuecomment-1978955833
5. Standard tools for processing experimental data (e.g. Mantid) will apply the $(kf/ki)^{-1}$ prefactor to the measured intensities. I will update the document to explain this.

PR #242, opened 2024-03-03
Remove most unicode from public API

kbarros:
Replace keyword arguments:
 * `Δt` with `dt`
 * `λ` with `damping`

Using the old symbols will emit a deprecation warning, and will break in 0.6.


ddahlbom (commented):
Took a quick pass over this. Looks good to me.

You want to eliminate the omegas and and energy_epsilon at a later point?

Issue #190, opened 2023-11-08
Incorrect sign of magnetic moment! (Breaking change for v0.6)


kbarros:
Sunny currently defines:

```julia
magnetic_moment(sys::System, site) = sys.units.μB * sys.gs[site] * sys.dipoles[site]
```

It is conventional to take $g \approx 2$ positive (consistent with the [electron-spin g-factor](https://en.wikipedia.org/wiki/G-factor_(physics)#Electron_spin_g-factor)). In this case, the magnetic moment has the wrong sign! (Because electron charge is negative, magnetic moment should point opposite to the angular momentum dipole.) For now, a practical workaround is to reverse the sign of the external field in all simulations. In the future, we should reverse the direction of the magnetic moment. This will be a breaking change, so it should wait until Sunny 0.6 release.

Problem identified by Radu Coldea.

PR #247, opened 2024-03-12
Fix signs of magnetic moments and LL spin dynamics

kbarros:
This resolves https://github.com/SunnySuite/Sunny.jl/issues/190.


src/Integrators.jl
```diff
@@ -262,20 +262,20 @@ end
     λ = damping
 
     if iszero(λ)
-        @. Δs = -s × (- dt*∇E)
+        @. Δs = - s × (dt*∇E)
     else
-        @. Δs = -s × (- dt*∇E + ξ - dt*λ*(s × ∇E))
+        @. Δs = - s × (ξ + dt*∇E - dt*λ*(s × ∇E))
```
ddahlbom, line 267: I guess this must work, but if the sign of the gradient must effectively change, shouldn't it do so in both places to get appropriate damping? I.e., `- s × (ξ + dt*∇E - dt*λ*(s × ∇E))` -> `- s × (ξ + dt*∇E + dt*λ*(s × ∇E))`.

src/SpinWaveTheory/HamiltonianDipole.jl
```diff
@@ -20,7 +20,7 @@ function swt_hamiltonian_dipole!(H::Matrix{ComplexF64}, swt::SpinWaveTheory, q_r
     (; extfield, gs, units) = sys
     for i in 1:L
         B = units.μB * (gs[1, 1, 1, i]' * extfield[1, 1, 1, i]) 
-        B′ = dot(B, local_rotations[i][:, 3]) / 2 
+        B′ = - dot(B, local_rotations[i][:, 3]) / 2
```
ddahlbom, line 23: As we discussed, it's a bit strange to me too that this is necessary (as far as consistency with SU(N) Hamiltonian construction), but it seems to result in consistent results when tested on a chiral model.

ddahlbom (commented):
This looks good to me, especially as it seems everything is working. A couple questions/comments above.

kbarros:
The sign of the LL damping term was already correct prior to this PR. The only thing that this is changing with this PR is the time direction for the energy conserving dynamics. (There's also a sign change on this noise term for aesthetics, but this is statistically identical.)

src/Integrators.jl
kbarros: This PR changes the time direction for the energy conserving dynamics, but does not change the sign of damping, which should be `Δs ~ s × (s × ∇E)`. If we changed this sign, the dynamics would inject energy rather than damp energy. In other words, the damping behavior was already correct prior to this PR.

src/SpinWaveTheory/HamiltonianDipole.jl
kbarros: The lack of a minus sign in dipole-LSWT code (prior to this PR) also seemed strange to me. I had a phone call with Hao, and he explained that it's correct, and has to do with the way that the spin operator is encoded in bosons. Hao said the formula is something like: $\hat{S} \sim - b^\dagger b$, which is the origin of the minus sign appearing in the Hamiltonian construction.

PR #263, opened 2024-05-14
Support varying spin-S in dipole mode

kbarros:
This fixes two bugs:

1. Indexing in `spin_label` was previously incorrect.
2. LSWT in dipole mode was using the first spin as representative of all spins.

### DIFF
--- a/src/SpinWaveTheory/SpinWaveTheory.jl
+++ b/src/SpinWaveTheory/SpinWaveTheory.jl
@@ -182,15 +182,13 @@ end
 
 # Compute Stevens coefficients in the local reference frame
 function swt_data_dipole(sys::System{0}, obs)
-    N = sys.Ns[1]
-    S = (N-1)/2
-
     cs = StevensExpansion[]
     Rs = Mat3[]
     Vs = Mat5[]
 
     n_magnetic_atoms = natoms(sys.crystal)
     observables_localized = zeros(ComplexF64, 1, 3, num_observables(obs), n_magnetic_atoms)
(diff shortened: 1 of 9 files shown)


src/SpinWaveTheory/SpinWaveTheory.jl
```diff
@@ -227,18 +225,18 @@ function swt_data_dipole(sys::System{0}, obs)
...
         for c in eachindex(ints.pair)
-            (; isculled, bond, scalar, bilin, biquad, general) = ints.pair[c]
-            i, j = bond.i, bond.j
+            (; bond, scalar, bilin, biquad, general) = ints.pair[c]
+            (; i, j) = bond
 
             if !iszero(bilin)  # Leave zero if already zero
                 J = Mat3(bilin*I)
-                bilin = S * (Rs[i]' * J * Rs[j]) 
+                bilin = sqrt(S[i]*S[j]) * Rs[i]' * J * Rs[j]
```
Hao-Phys, line 233: I think it is a good idea to remove any `S`s in the definition of `billin` to support the varying-S spin wave theory.

Hao-Phys: Also, removing `S`s from the definition would help my implementation of the non-perturbative calculation. It's imperative to maintain the vertex function's order in terms of `S`. Currently, I have to divide everything by `S`, which can be somewhat counterintuitive.

kbarros: Yes I agree.

PR #265, opened 2024-05-22
SingleQ integration

Lazersmoke:
This PR further integrates the new SingleQ solvers by doing several things:

- All `intensities_*_SingleQ` functions are gone
- The internal interface for Broadened Single-Q is no longer band-resolved. This is to make it in line with the existing non-band-resolved broadened spin wave code. The alternative design choice is to make both of these band resolved (and would result in duplicated `intensities_broadened` code between them to deal with the fact that single-q has 3x as many bands).
- All `intensities_*` functions for spin wave that can be made to work generically for both ordinary and Single-Q spin wave now support both. Specifically, this includes the most experimentally relevant one: `intensities_bin_multisample` as well as all other broadened modes.
- The remaining `intensities_*` function, which is `intensities_bands`, can't be made totally generic because of the aforementioned 3x as many bands in Single-Q mode. Instead, this is now decided by dynamic dispatch (a Julia win!)


kbarros:
I have cherry-picked the two commits from this PR into https://github.com/SunnySuite/Sunny.jl/pull/230, which effectively merges this PR.

PR #269, opened 2024-05-28
Document magnetic moment sign error

kbarros:
A stopgap until https://github.com/SunnySuite/Sunny.jl/pull/247 is merged in Sunny 0.6.

### DIFF
--- a/src/System/Interactions.jl
+++ b/src/System/Interactions.jl
@@ -66,7 +66,7 @@ end
 """
     enable_dipole_dipole!(sys::System)
 
-Enables long-range interactions between magnetic moment dipoles,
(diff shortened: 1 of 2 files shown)


src/System/Interactions.jl
```diff
@@ -66,18 +66,17 @@ end
 """
     enable_dipole_dipole!(sys::System)
 
-Enables long-range dipole-dipole interactions,
+Enables long-range interactions between magnetic moment dipoles,
```
ddahlbom, line 69: Would "magnetic dipole moments" be more conventional?

ddahlbom (commented):
Looks like a sensible approach to me.

PR #272, opened 2024-06-06
Classical intensities as density

ddahlbom:
This PR
- Divides the output of `intensities_interpolated` by `sc.Δω`so that the returned values are densities. If `sc.Δω=NaN` (because `sc` is an instant structure factor) divides by 1 instead.
- Updates tests correspondingly.
- Updates `colorrange`s of heatmap plots in examples.
- Pre-plans additional FFTs for new correlation scheme and eliminates an allocation.


examples/04_GSD_FeI2.jl
```diff
@@ -283,11 +283,11 @@ ax_top = Axis(fig[1,1],ylabel = "meV",xticklabelrotation=π/8,xticklabelsize=12;
 ax_bottom = Axis(fig[2,1],ylabel = "meV",xticks = (markers, string.(points)),xticklabelrotation=π/8,xticklabelsize=12)
 
 heatmap!(ax_top,1:size(is_interpolated,1), ωs, is_interpolated;
-    colorrange=(0.0,0.07),
+    colorrange=(0.0, 1.0),
 )
 
 heatmap!(ax_bottom,1:size(is_binned,1), ωs, is_binned;
-    colorrange=(0.0,0.05),
+    colorrange=(0.0, 1.5e-6),
```
kbarros, line 290: Did we accidentally introduce a Δω factor into binned intensities? These are already integrated quantities, so no Δω should appear, and the colorrange should not need to change.

ddahlbom: This was the figure that was already "broken" (i.e., appeared blank) before anything was added by this PR. In other words, this change was essentially unrelated to the rest of the PR.

@Lazersmoke I don't think any of the binning functions call `intensities_interpolated`, so they should not be affected by this PR, and all the binning tests still pass. But the binned results in the FeI2 tutorial appear appear like they already needed some plotting tweaks. See binned plot [here](https://sunnysuite.github.io/Sunny.jl/dev/examples/04_GSD_FeI2.html). The change here just tweaks the plotting limits so that the results show up.

kbarros:
Looks good to me, pending Sam's comments on binned intensities. Thanks David.

Lazersmoke:
Hi David, thanks for writing this up.

I think the better way to make this change is to remove the 1/sqrt(n all omega) from the time fft normalization factor. The complete end to end calculation in the time direction is then: 

1. For each time offset t, sum up the `|t-T|` estimates (by fft method), and divide by that number of estimates. This results in a density already! (It's a number that does not go down when the bin size is decreased) 
2. For interpolation, just return the density (no factor) 
3. For binning, calculate and include the factor (this is done at the end of intensities_binned already, which is why there's a bug happening in the docs)

Lazersmoke:
(per the discussion around #246 , the overall color scale should have not changed during that, but it turns out it actually did because the aforementioned inclusion of the factor in step 3 happened when that was merged)

Lazersmoke:
See #273

examples/04_GSD_FeI2.jl
kbarros: Per discussions on Slack, the binning intensity scale may have been broken in https://github.com/SunnySuite/Sunny.jl/pull/271, but is now fixed through the commits in https://github.com/SunnySuite/Sunny.jl/pull/273, which have been merged into this PR branch.

PR #277, opened 2024-06-15
Support loading magnetic structure from MCIF

kbarros:
This probably needs lots of testing. In particular, I am not sure whether BNS vs OG setting is relevant.
https://www.cryst.ehu.es/cryst/magnext_help.html


kbarros:
Daniel Pajerowski helped with testing, and suggested a nice way to load crystal cells from mCIF files.

PR #284, opened 2024-07-03
Change Units.theory conventions

kbarros:
* Ensure that the Bohr magneton $μ_B$ is positive, even with "theory" units. To include a Zeeman coupling of the form $- S ⋅ B$, it is now required to select negative `g = -1` in addition to selecting `Units.theory`.
* Disallow dipole-dipole interactions with `Units.theory` -- such an interaction doesn't make sense once the energy and length scale units become ambiguous.
* Modify `magnetic_moment` so that its return value is in units of $μ_B$.

### DIFF
--- a/test/test_ewald.jl
+++ b/test/test_ewald.jl
@@ -18,13 +18,14 @@
     positions = [[0,0,0]]
     cryst = Crystal(latvecs, positions)
     infos = [SpinInfo(1, S=1, g=1)]
-    sys = System(cryst, (1,1,1), infos, :dipole; units=Units.theory)
+    units = Sunny.PhysicalConsts(; μ0=1, μB=1)
+    sys = System(cryst, (1,1,1), infos, :dipole; units)
     enable_dipole_dipole!(sys)
     @test ewalder_energy(sys) ≈ -1/6
     @test isapprox(energy(sys), -1/6; atol=1e-13)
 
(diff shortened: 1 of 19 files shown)


test/test_ewald.jl
```diff
@@ -18,13 +18,14 @@
     positions = [[0,0,0]]
     cryst = Crystal(latvecs, positions)
     infos = [SpinInfo(1, S=1, g=1)]
-    sys = System(cryst, (1,1,1), infos, :dipole; units=Units.theory)
+    units = Sunny.PhysicalConsts(; μ0=1, μB=1)
```
ddahlbom, line 21: I wonder if this should be the default for theory units (permeability of 1) rather than simply disallowing dipole-dipole.

ddahlbom (commented):
This all looks good to me.

My only question is about whether the default theory units should disallow dipole-dipole. The code in `test_ewald.jl` picks units that seem natural for a theory calculation.

test/test_ewald.jl
kbarros: As far as I know, the only current use case would be for testing, or for theoretical models designed specifically to study dipole-dipole effects. In the latter case, the theorist will likely want a tunable dipole-dipole interaction strength, not some specific arbitrary one.

PR #285, opened 2024-07-10
Flexible Units

kbarros:
Allow more flexibility in the system of `Units`. Given some specific energy and length scale, e.g. `units = Units(:meV, :angstrom)`, the returned object will provide various physical constants. For example, `units` provides `THz`, `meV`, `K`, and `T` in multiples of the base energy unit. For kelvin and tesla, conversions are employed using the Botzmann constant and Bohr magneton, respectively. The function `set_external_field!` is deprecated in favor of `set_field!`, which now expects the magnetic field scaled by the Bohr magneton, which has units of energy. The function `enable_dipole_dipole!` now expects a vacuum permeability constant $μ_0 μ_B^2$ with dimensions of length³-energy, and this can be obtained via `units.vacuum_permeability`.

Most existing Sunny scripts should continue to work, but deprecation warnings will guide users toward the new interface.


ddahlbom (approved):
As I indicated in our Slack conversation, this all looks great to me. I think it's a much nicer approach than we had before.

One minor question I have is whether its actually `vacuum_permeability` in the units. Isn't it vacuum permeability times the square of the Bohr magneton?

kbarros:
> One minor question I have is whether its actually vacuum_permeability in the units. Isn't it vacuum permeability times the square of the Bohr magneton?

Yes, that's true. If we wanted to be more complete, we could hypothetically perform the following renamings of the Units energy constants to include the conversion factors:

```
THz -> h_THz   # Planck constant
K -> kB_K      # Boltzmann constant
T -> μB_T      # Bohr magneton
```

But that is sort of unnecessarily verbose. So colloquially, a physicist might say "work in units where $h = k_B = μ_B= 1$". Given that `units.T` actually means $μ_B T$, I think it's consistent to say that `units.vacuum_permeability` (namely $μ_0$) actually means $μ_0 μ_B^2$.

ddahlbom:
Sounds good. I have no objections, just wanted to make sure I'm following.

PR #297, opened 2024-08-23
Small cleanups to SampledCorrelations

kbarros:
Clarify some of the logic around `SampledCorrelations`.

Also, `thermal_prefactor` now correctly goes to 1/2 when ω = kT = 0.


ddahlbom (approved):
Looks good to me. Thank you.

As you say, we do probably want further refactoring later. (intensities_rounded is isolated and type-stable, but it's annoying that intensities itself is not type stable.)

PR #304, opened 2024-08-27
Rewrite of Spiral SWT

kbarros:
With this PR, spiral LSWT calculations employ exchange interactions expressed in Fourier space, $J(q)$. This allows to support long-range dipole-dipole interactions. The new method was developed and implemented by Alin Niraula and Xiaojian Bai.

This PR also includes a bug fix: the propagation wavevector $k$ is now categorized into one of three "special" cases via the new function `spiral_propagation_case`. This corrects the behavior for FM and AFM structures (interpreted as generalized spirals), which involve $q$ with integer and half-integer components, respectively.

Performance benchmarks are roughly unchanged after the switch from real- to Fourier-space exchange. A simple test case is the `powder_average` call in SpinW tutorial 08 ($\sqrt{3}\times\sqrt{3}$ phase of the Kagome lattice).


kbarros:
[Tests.zip](https://github.com/user-attachments/files/17399195/Tests.zip)

Two tests from Alin: (1) Canted AFM and (2) DM with single-ion anisotropy. There is no test on single-ion anisotropy because any small amount deforms the spiral phase.

PR #306, opened 2024-08-29
Uncomment contributed build commands

ddahlbom:
Reintroduces the contributed docs to build.


kbarros (approved):
Looks good to me, it's probably reasonable that Contributed comes after SpinW ports.

PR #311, opened 2024-09-01
Fix intensity scale for SampledCorrelations with reshaped system

kbarros:
Add a test that covers the case. Various other aesthetic changes are included.


ddahlbom (approved):
This all looks good to me.

PR #317, opened 2024-09-12
Avoid crashes when reporting symmetry-allowed interactions

### LINKED ISSUE #260: Assertion error in `stevens_basis_for_symmetry_allowed_anisotropies` triggered by certain crystals
Sometimes the analysis of allowed onsite anisotropies fails, for example when calling `print_symmetry_table`. Specifically, one may get `AssertionError: norm(imag(B)) < 1.0e-12`, which is thrown [here](https://github.com/SunnySuite/Sunny.jl/blob/1c7cc3cd28ba0afe1ee4f7a43daf711f1594dd5c/src/Symmetry/AllowedAnisotropy.jl#L92).

A minimal example which produces this error is given by an FCC lattice with rhombohedral distortion:
```
a = 6.22
distortion = 0.15
latvecs = lattice_vectors(a, a, a, 90+distortion, 90+distortion, 90+distortion)
positions = Sunny.fcc_crystal().positions
crystal = Crystal(latvecs, positions; types = ["A", "B", "B", "B"])
print_symmetry_table(crystal, 7.0)
```
The values of `distortion` for which this happens are hard to predict. For example, `distortion=5.0` works fine, as does `distortion=1`. `distortion=1.1` fails.

@kbarros checked that the rotation matrix derived from the lattice vectors is consistently orthogonal to high precision, as is the unitary constructed from that rotation matrix.

kbarros:
Avoid crashes in `print_symmetry_table` when symmetry-allowed expansion coefficients have a wide range of magnitudes. Fixes #260.


src/Symmetry/AllowedAnisotropy.jl
```diff
@@ -61,38 +61,21 @@ function basis_for_symmetry_allowed_anisotropies(cryst::Crystal, i::Int; k::Int,
...
+        transform_spherical_to_stevens_coefficients(k, c)
+    end
 
-    # Transform each column c to coefficients b that satisfy bᵀ 𝒪 = cᵀ T
-    B = [transform_spherical_to_stevens_coefficients(k, c) for c in eachcol(C)]
+    # If 𝒜 is symmetry-allowed, then its Hermitian and anti-Hermitian parts are
+    # independently symmetry-allowed. These are given by the real and imaginary
+    # parts of B. Create a real matrix with these two parts, and eliminate
+    # linearly-dependent vectors from the column space.
+    B = colspace(hcat(real(B), imag(B)); atol=1e-12)
```
ddahlbom, line 75: I guess this in some sense eliminates the imaginary part by construction in a slightly different way. It's not obvious to me what the effect of doing it this way is, but it seems to work.

ddahlbom (approved):
Per our discussion on slack, this looks good to me.

src/Symmetry/AllowedAnisotropy.jl
kbarros: In the previous code, we were using an SVD to reduce the spanned vector space, and claiming "the result must be real". Although that seemed to be true in practice, it's not clear to me that SVD actually gives that guarantee.

In this PR, we're doing something more careful. The columns of B are complex vectors that span the space of symmetry-allowed operators, with no restriction on Hermiticity. However, we wish to apply an additional constraint that we only care about symmetry-allowed operators that are Hermitian. If we just take `real(B)`, we'd get symmetry-allowed operators that are Hermitian, but we're not guaranteed to get _all_ of them. Similarly, `imag(B)` will give a different space of symmetry-allowed operators, and if you multiply by `im`, then they are Hermitian also. I suspect that the combined columns `hcat(real(B), imag(B))` spans the full space of symmetry-allowed Hermitian operators. Then the call to `colspace` cleans up the redundancy.

PR #324, opened 2024-10-10
Entangled units updates

ddahlbom:
Fixes several bugs, including:
    - Error is `Base.show`
    - Error in initializing an external field from existing system
    - Error in updating external field
    - Sign convention for external field

Adds:
    - Dispersion tests for dimer model
    - Intensity golden test for dimer model using classical dynamics
    - Additional function aliasing (`clone_system`, `dispersion`, `resize_supercell`)


src/EntangledUnits/EntangledUnits.jl
```diff
@@ -254,6 +258,13 @@ function entangle_system(sys::System{M}, units) where M
     spin_infos = [i => Moment(s=(N-1)/2, g=1.0) for (i, N) in enumerate(Ns_contracted)]  # TODO: Decisions about g-factor 
     sys_entangled = System(contracted_crystal, spin_infos, :SUN; dims)
 
+    # Transfer rng from origin system to entangled system
+    sys_entangled.rng.s0 = sys.rng.s0
+    sys_entangled.rng.s1 = sys.rng.s1
+    sys_entangled.rng.s2 = sys.rng.s2
+    sys_entangled.rng.s3 = sys.rng.s3
+    sys_entangled.rng.s4 = sys.rng.s4
```
kbarros, line 266: Can this be replaced with `sys_entangled.rng = copy(sys.rng)`? The fields of an `rng` are private.

kbarros (commented):
Overall looks good to me.

src/EntangledUnits/EntangledUnits.jl
ddahlbom: It's a workaround for the fact that the `rng` field of a `System` is `const` -- admittedly a very ugly and brittle workaround. I can think of other ways of doing it, but they're all ugly in similar ways.

I'll give it some thought.

ddahlbom: Per your suggestion, used `Base.copy!` instead of accessing private fields.

ddahlbom:
In response to a request from Daniel Pajerowski, added `Sunny.coherent_state_from_dipoles`, a function that constructs a coherent state for an entangled unit based on a set of dipoles for each atom inside the entangled unit. These dipoles are all assumed to be fully polarized.

It is not possible to write a well-defined `set_dipole!` function for entangled systems, since all dipoles within a unit must be set at once. This function offers another path to initialize an entangled system in a classical dipolar state.

Note that it would be possible to write a batch `set_dipoles!` function for an entangled system, and perhaps this should be added. It would also be possible to set the state of entangled system from the coherent states of an SU(N) unentangled system, which could take the form of a `set_coherents!` function.

PR #326, opened 2024-10-30
Fix reference data for entangled units "dimer" test on Julia 1.11

kbarros:


ddahlbom (approved):
Thanks for doing this. (I would have been happy to. I should've opened something earlier.)

This looks good to me. I see the same numbers on my end for 1.11. I added a few `@test` annotations that I had forgotten.

PR #332, opened 2024-11-12
Add KPM to README and update links

kbarros:


ddahlbom (approved):
These look like good changes to me.

Issue #233, opened 2024-02-04
Deploy docs to separate repo


kbarros:
Our docs images are high res and quite large. Hosting docs for all releases will be quite large -- I could imagine this growing to ~ GB, which will slow down clones of Sunny. We could follow Meshes.jl and create a separate repo to store the gh-pages for docs deployment:
https://github.com/JuliaGeometry/Meshes.jl/blob/master/docs/make.jl

kbarros:
This is now described in the Documenter docs: https://documenter.juliadocs.org/stable/man/hosting/#Out-of-repo-deployment

kbarros:
Some more motivation. As of Sunny 0.7.3, cloning the `main` branch only requires 8.2 MB of space:
```
git clone --single-branch -b main git@github.com:SunnySuite/Sunny.jl.git
du -sh Sunny.jl
# 8.2M	Sunny.jl
```

A full clone downloads the full documentation in the `gh-pages` branch, which includes many Sunny releases. This takes 131 MB of space:
```
git clone git@github.com:SunnySuite/Sunny.jl.git
du -sh Sunny.jl
# 131M	Sunny.jl
```

PR #334, opened 2024-11-15
Default seed for a `System` is generated from the task-local RNG

kbarros:
The default `System` random number seed is now `seed=rand(UInt)`. The motivation is that calling `Random.seed!` prior to `System` construction will fix `System.rng` in a deterministic way. This may be helpful for writing reproducible Sunny scripts that sometimes _also_ call Julia's task-local RNG.

An underlying motivation is to make the doc builds exactly reproducible. This will hopefully allow for more compression on the `gh-pages` branch, which already requires 160+ MB to store docs images, with repetition on every Sunny release. See also https://github.com/SunnySuite/Sunny.jl/issues/233#issuecomment-2476980802.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -37,6 +37,7 @@ function build_examples(example_sources, destdir)
             """
             # Download this example as [Julia file]($assetsdir/scripts/$name.jl) or [Jupyter notebook]($assetsdir/notebooks/$name.ipynb).
 
+            import Random; Random.seed!(0); #hide
             """ * str
         end
         # Write to `src/$destpath/$name.md`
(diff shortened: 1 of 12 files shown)


docs/make.jl
```diff
@@ -37,6 +37,7 @@ function build_examples(example_sources, destdir)
             """
             # Download this example as [Julia file]($assetsdir/scripts/$name.jl) or [Jupyter notebook]($assetsdir/notebooks/$name.ipynb).
 
+            import Random; Random.seed!(0); #hide
             """ * str
```
ddahlbom, line 41: Nice

ddahlbom (approved):
I think this is a good change. Thanks, Kip.

PR #337, opened 2024-11-17
Update the documentation landing page

kbarros:
The docs landing page better describes what Sunny is, and provides links to its various features.

Within the Library Reference page, URLs for function docstrings now omit type information.


ddahlbom (commented):
The changes to the landing page look good to me. Concise and effective.

I noticed that some of the early advanced MC efforts are not highlighted. Is the idea that they need to be updated to the current level of polish before getting clearly called out?

PR #339, opened 2024-12-03
Replace KPM with Lanczos

kbarros:
The Lanczos algorithm promises these benefits relative to KPM:

 - Higher accuracy at a fixed number of iterations (i.e., at a fixed Krylov subspace dimension).
 - Does not require an initialization phase that bounds the spectrum prior to polynomial expansion.
 - Ability to detect instabilities (negative eigenvalues) in the dynamical matrix.
 - Allows for a sharp cutoff between positive and negative eigenvalues near zero. ~~Whereas KPM is susceptible to a loss of intensity near Goldstone modes, Lanczos seems to avoid this problem.~~ (Unfortunately, despite my initial hopes, intensity is still lost via various mechanisms. See examples below.)

As currently implemented, Lanczos is slower than KPM _per iteration_. But fewer iterations are required, so Lanczos still seems favorable in overall speed. And future optimizations may be possible, e.g., operating on multiple right-hand side vectors in parallel.

A possible disadvantage of Lanczos is the accumulation of numerical roundoff error (loss of orthogonality in basis vectors for the Krylov space), and the effects of this are harder to quantify. See https://arxiv.org/abs/2410.11090 for a modern mathematical review of the Lanczos method. A backward stability analysis, in Sec. 4, suggests that the instabilities of Lanczos are not catastrophic for the desired observables. Empirically, however, Lanczos can still exhibit severe artifacts in estimating the intensities of bands near zero energy (quantified in examples below). Note, however, that these are cases where KPM also fails.

Another consideration is that Lanczos requires direct diagonalization of a tridiagonal matrix. After $k$ Lanczos iterations, this cost should scale like $\mathcal{O}(k^2)$, which may dominate if $k$ sufficiently large.

### DIFF
--- a/docs/src/structure-factor.md
+++ b/docs/src/structure-factor.md
@@ -374,12 +374,12 @@ calculated to leading order in inverse powers of the quantum spin-$s$. For
 systems constructed with `mode = :SUN`, Sunny automatically switches to a
 multi-flavor boson variant of spin wave theory, which captures more single-ion
 physics. Use [`SpinWaveTheorySpiral`](@ref) to study generalized spiral phases,
-which allow for an incommensurate propagation wavevector. The experimental
-module [`SpinWaveTheoryKPM`](@ref) implements [spin wave calculations using the
-kernel polynomial method](https://arxiv.org/abs/2312.0834). In the KPM approach,
-the computational cost scales linearly in the magnetic cell size. It can be
-useful for studying systems with large magnetic cells include systems with
-long-wavelength structures, or systems with quenched chemical disorder.
+which allow for an incommensurate propagation wavevector. The module
+[`SpinWaveTheoryKPM`](@ref) uses [iterated matrix-vector
+multiplication](https://arxiv.org/abs/2312.0834) to achieve linear-scaling cost
+in the magnetic cell size. It can be useful for studying systems with large
+magnetic cells include systems with long-wavelength structures, or systems with
+quenched chemical disorder.
 
 
 ## Calculations with classical spin dynamics
--- a/docs/src/versions.md
+++ b/docs/src/versions.md
@@ -7,6 +7,8 @@
   wave theory. Empty sites are modeled using bosons that do not excite.
 * Fix correctness of [`suggest_magnetic_supercell`](@ref) when multiple
   wavevectors are provided.
+* The default implementation of [`SpinWaveTheoryKPM`](@ref) now uses Lanczos for
+  higher accuracy.
 
 ## v0.7.5
 (Jan 20, 2025)
--- a/docs/src/why.md
+++ b/docs/src/why.md
@@ -38,7 +38,7 @@ Feature highlights include:
   on [very large magnetic cells](@ref SpinWaveTheoryKPM). The [FeI₂
   example](@ref "3. Multi-flavor spin wave simulations of FeI₂") showcases LSWT
   with multi-flavor bosons. The [disordered system example](@ref "9. Disordered
-  system with KPM") demonstrates acceleration for large system sizes.
+  systems") demonstrates acceleration for large system sizes.
 - [Dipole-dipole interactions](@ref enable_dipole_dipole!) with full Ewald
   summation, as illustrated in the [pyrochlore LSWT example](@ref "7. Long-range
   dipole interactions"). Dipole-dipole interactions in classical dynamics are
--- a/examples/09_Disorder_KPM.jl
+++ b/examples/09_Disorder_KPM.jl
@@ -47,12 +47,13 @@ plot_intensities(res)
 
 sys_inhom = to_inhomogeneous(repeat_periodically(sys, (10, 10, 1)))
 
-# Use [`symmetry_equivalent_bonds`](@ref) to iterate over all nearest neighbor
-# bonds of the inhomogeneous system. Modify each AFM exchange with a noise term
-# that has variance of 1/3. The newly minimized energy configuration allows for
-# long wavelength modulations on top of the original 120° order.
+# Iterate over all [`symmetry_equivalent_bonds`](@ref) for nearest neighbor
+# sites in the inhomogeneous system. Set the AFM exchange to include a noise
+# term that has a standard deviation of 1/3. The newly minimized energy
+# configuration allows for long wavelength modulations on top of the original
+# 120° order.
 
-for (site1, site2, offset) in symmetry_equivalent_bonds(sys_inhom, Bond(1,1,[1,0,0]))
+for (site1, site2, offset) in symmetry_equivalent_bonds(sys_inhom, Bond(1, 1, [1, 0, 0]))
     noise = randn()/3
     set_exchange_at!(sys_inhom, 1.0 + noise, site1, site2; offset)
 end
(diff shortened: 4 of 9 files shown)


kbarros:
Here is a minimal example where Lanczos fails to get the correct intensity at the small energy mode:

```jl
cryst = Sunny.bcc_crystal()
sys = System(cryst, [1 => Moment(; s=1, g=2)], :dipole; dims=(1, 1, 1))
sys = to_inhomogeneous(sys)
set_exchange_at!(sys, -1.0, (1, 1, 1, 1), (1, 1, 1, 2); offset=[0, 0, 0]) 

polarize_spins!(sys, [0, 0, 1])
energy_per_site(sys)

kernel = lorentzian(fwhm=0.2)
energies = range(0, 6, 1000)
qs = [[0.35, 0.35, 0]]

measure = ssf_custom((q, ssf) -> real(ssf[1, 1]), sys) # Sˣˣ(q,ω) only

swt = SpinWaveTheoryKPM(sys; measure, lanczos_iters=4, tol=1, regularization=1e-8)
res = intensities(swt, qs; energies, kernel)

swt0 = SpinWaveTheory(sys; measure, regularization=1e-8)
res0 = intensities(swt0, qs; energies, kernel)

lines(vec(res.data))
lines!(vec(res0.data))
```

image

In this particular case, the magnetic cell consists of two sites and a single bond. The dynamical matrix has size 4×4. The two modes should energies of 2 and 0, slightly modified by the default regularization of $ϵ = 10^{-8}$,

```jl
D = Sunny.dynamical_matrix(swt.swt, Sunny.Vec3(only(qs)))
It = Hermitian(diagm([fill(1.0, 2); fill(-1.0, 2)]))
eigvals(It * D)
# ±(2 + 2ϵ), ±(2ϵ + √2ϵ^2)
```

However, there is an inversion ("particle-hole") symmetry in this problem that Lanczos doesn't exactly respect. Specifically, because $D_q = D_{-q}$, and similarly for the right-hand side vector representing the observable, all the diagonal elements $(α_1, ..., α_k)$ of the Lanczos tridiagonal matrix should be exactly zero. In practice, however, once this symmetry becomes broken, the Lanczos iterations allow significant "mixing" between eigenvectors in the negative and positive parts of the spectrum near zero. In practice, this mixing causes huge error in the estimated intensities near zero energy.

A workaround for this specific problem could be to modify the Lanczos iterations to exactly respect inversion symmetry, e.g., following the algorithm in Fig. 4 of Grüning et al, _Implementation and testing..._, [arXiv:1102.3909](https://arxiv.org/abs/1102.3909v1). However, many practical problems in LSWT will _not_ have this exact inversion symmetry, and a more general solution is still needed.

One idea is to double the dimension of the matrix size by working with both wavevectors $\pm q$ simultaneously. Then, in principle, the symmetrized Lanczos version would always apply, and the positive and negative parts of the spectrum become redundant. However, this still leaves the possibility of mixing between the eigenvectors of the positive part of the spectrum, and it's not clear there is a win.

kbarros:
Here is the code for a more realistic scenario where Lanczos completely fails. This builds a 3x3 system, nearest neighbor ferromagnetic interactions, where 40% sites are replaced by vacancies.

```jl
import Random

function create_dilute_sys(n, x; s=1, mode=:dipole)
    a = b = 1.0
    c = 10.0
    latvecs = lattice_vectors(a, b, c, 90, 90, 90)
    positions = [[0, 0, 0]]
    cryst = Crystal(latvecs, positions)
    J₁ = 1.0
    sys = System(cryst, [1 => Moment(; s, g=2)], mode; dims=(n, n, 1))
    set_exchange!(sys, J₁, Bond(1, 1, [1, 0, 0]))

    sys_inhom = to_inhomogeneous(sys)
    nvacancies = Int(round(n*n*x))

    vacancy_sites = sample(eachsite(sys_inhom), nvacancies; replace=false)
    for site in vacancy_sites
        Sunny.set_vacancy_at!(sys_inhom, site)
    end

    randomize_spins!(sys_inhom) 
    minimize_energy!(sys_inhom; maxiters=10_000)
    return sys_inhom, cryst
end

kernel = lorentzian(fwhm=0.2)
energies = range(0, 6, 1000)
qs = [[0.35, 0.35, 0]]

Random.seed!(2)
n = 3
x = 0.4

sys, cryst = create_dilute_sys(n, x)

measure = ssf_custom((q, ssf) -> real(ssf[1, 1]), sys)

swt = SpinWaveTheoryKPM(sys; measure, lanczos_iters=2, tol=1, regularization=1e-8)
res_Lanczos = intensities(swt, qs; energies, kernel)

swt_lswt = SpinWaveTheory(sys; measure, regularization=1e-8)
res_LSWT = intensities(swt_lswt, qs; energies, kernel)

lines(vec(res_Lanczos.data), label="Lanczos")
lines!(vec(res_LSWT.data), label="Reference")
axislegend()
```

image

To emphasize the failure, just two Lanczos iterations are being executed, yielding a single estimated band. The discrete Lanczos band is placed around the energy of 0.9, and has very little intensity compared the band near zero energy.

kbarros:
Here's another modification of Harry's site dilution test. This one uses ferromagnetic interactions with an applied field, to avoid zero energy modes. It seems clear that the improvements in Lanczos accuracy make up for the slower time-per-iteration. And note that the Lanczos implementation could be further optimized.

```jl
using Sunny
import StatsBase, Random

function create_dilute_sys(n, x; s=1, mode=:dipole)
    a = b = 1.0
    c = 10.0
    latvecs = lattice_vectors(a, b, c, 90, 90, 90)
    positions = [[0, 0, 0]]
    cryst = Crystal(latvecs, positions)
    J₁ = -1.0
    sys = System(cryst, [1 => Moment(; s, g=2)], mode; dims=(n, n, 1))
    set_exchange!(sys, J₁, Bond(1, 1, [1, 0, 0]))
    set_field!(sys, [0, 0, 0.2]) # Ensure some small gap

    sys_inhom = to_inhomogeneous(sys)
    nvacancies = Int(round(n*n*x))

    vacancy_sites = StatsBase.sample(eachsite(sys_inhom), nvacancies; replace=false)
    for site in vacancy_sites
        Sunny.set_vacancy_at!(sys_inhom, site)
    end

    randomize_spins!(sys_inhom) 
    minimize_energy!(sys_inhom; maxiters=10_000)
    return sys_inhom, cryst
end

kernel = lorentzian(fwhm=0.2)
energies = range(0, 8, 1000)
qs = [[0.35, 0.35, 0]]

Random.seed!(0)
n = 12
x = 0.4

sys, cryst = create_dilute_sys(n, x)

measure = ssf_custom((q, ssf) -> real(ssf[1, 1]), sys)
regularization = 1e-8

using GLMakie, ColorSchemes

fig = Figure()

for (panel, method) in enumerate((:lanczos, :kpm))
    swt_lswt = SpinWaveTheory(sys; measure, regularization)
    res = intensities(swt_lswt, qs; energies, kernel)

    ax = Axis(fig[panel, 1])
    lines!(ax, energies, vec(res.data), label="Reference", color=:black, linestyle=:dash, linewidth=3)

    colors = reverse(resample(ColorSchemes.buda50, length(all_niters)+1))[2:end]
    all_niters = [32, 64]

    for (niters, color) in zip(all_niters, colors)
        swt = SpinWaveTheoryKPM(sys; measure, niters, regularization, method)
        res = intensities(swt, qs; energies, kernel)
        lines!(ax, energies, vec(res.data); color, label=":$method, niters=$niters")
    end

    axislegend(ax)
end

fig
```

image

Issue #60, opened 2023-02-17
Consistency in coefficients for Stevens expansion


kbarros:
The [`print_site`](https://sunnysuite.github.io/Sunny.jl/stable/library.html#Sunny.print_site) function accepts an optional parameter `R` that sets a reference frame for the Stevens expansion. Currently, the coefficients ($c_1$, $c_2$, ...) are calculated from scratch for each new reference frame. Instead, we could give the coefficients a consistent meaning by first calculating the Stevens expansion for `R=I` and then rotating the expansion symbolically.

We could also maintain consistency between two sites by allowing to select a reference atom `i_ref`, which would be analogous to the `b_ref` parameter in [`print_bond`](https://sunnysuite.github.io/Sunny.jl/stable/library.html#Sunny.print_bond).

PR #342, opened 2024-12-09
Consistency in expansion of anisotropy

kbarros:
Allow `print_site` to accept an optional reference atom `i_ref`. Keeping this "reference atom" fixed between calls to `print_site` will ensure a consistent meaning of the free expansion parameters. Note that `i_ref` is analogous to the "reference bond" parameter `b_ref` that is already accepted by `print_bond`. To achieve more uniformity between `print_site` and `print_bond`, this PR changes the default value of `b_ref` to `b` (before, it required selecting an element from `reference_bonds`).

As before, `print_site` accepts an optional rotation `R` to globally change the coordinate system. This functionality now works better -- in particular, the expression of the allowed ``g`` becomes cleaner.

Resolves https://github.com/SunnySuite/Sunny.jl/issues/60.


ext/PlottingExt/ViewCrystal.jl
```diff
@@ -60,6 +60,50 @@ function propagate_reference_bond_for_cell(cryst, b_ref)
     return reduce(vcat, found)
 end
 
+# Get the quadratic anisotropy as a 3×3 exchange matrix for atom `i` in the
+# chemical cell.
+function quadratic_anisotropy(sys, i)
```
ddahlbom, line 65: I wonder if maybe an asterisk or warning should be given (in the display) if c4 or c6 is non-zero?

My worry is just that one of the selling points of Sunny has been its support for completely general anisotropies. This potentially truncated representation of anisotropies might be misleading to someone who wants to use the advanced anisotropy features and specifically looks at the viewer to check things.

ddahlbom (approved):
I do not have too much intelligent to say about this. It just reminded me of how much has gone into the symmetry analysis, and it's interesting to poke through the code. It looks very nice.

I added one comment above about whether we should warn if the anisotropy is "truncated" in the display.

Eventually it might be nice to add a visualization of anisotropies when they have been assigned (e.g., as a heatmap on a sphere). It doesn't seem urgent, but it would be fun to do at some point.

ext/PlottingExt/ViewCrystal.jl
```diff
@@ -287,7 +331,17 @@ function label_atoms(cryst; ismagnetic)
...
         if ismagnetic
-            # TODO: Show onsite couplings?
+            if isnothing(sys)
+                # See similar logic in print_site()
+                refatoms = [b.i for b in Sunny.reference_bonds(cryst, 0.0)]
+                i_ref = Sunny.findfirstval(i_ref -> Sunny.is_related_by_symmetry(cryst, i, i_ref), refatoms)
+                R_site = Sunny.rotation_between_sites(cryst, i, i_ref)
+                push!(ret, Sunny.allowed_g_tensor_string(cryst, i_ref; R_site, prefix="Aniso: ", digits=8, atol=1e-12))
+            else
+                ansio = quadratic_anisotropy(sys, i)
```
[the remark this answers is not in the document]
kbarros: Good catch, fixed

ext/PlottingExt/ViewCrystal.jl
kbarros: Good idea. Now `view_crystal(sys)` will display the text _"+ higher order terms"_ if there are any terms beyond the quadratic level. This behavior has been added for both pair couplings (on bonds) and anisotropy (on sites).

PR #346, opened 2024-12-19
Write examples/extra folder as Jupyter notebooks

quantumsteve:
This updates comments from `#` to `##` in the examples/extra folder so that they can be written as Jupyter notebooks.

### DIFF
--- a/examples/extra/Advanced_MC/PT_WHAM_ising2d.jl
+++ b/examples/extra/Advanced_MC/PT_WHAM_ising2d.jl
@@ -34,10 +34,10 @@ Sunny.step_ensemble!(PT, n_therm, exch_interval)
 
 # Start PT simulation
 for _ in 1:n_measure
-    # Run some sweeps and replica exchanges
+    ## Run some sweeps and replica exchanges
     Sunny.step_ensemble!(PT, measure_interval, exch_interval)
 
-    # Measurements - assuming LocalSampler used
+    ## Measurements - assuming LocalSampler used
     for (j, sampler) in enumerate(PT.samplers)
         E_hists[j][sampler.ΔE] += 1
     end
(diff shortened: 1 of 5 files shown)


kbarros:
Big picture comment -- files `examples/extra` are not really intended as tutorials for end users. Perhaps they should be better hidden, rather than exposed via notebooks?

* An exception is the parallel tempering example -- many users will want this, so it is is a good candidate to be cleaned up and included as a full-blown tutorial in the main docs. Probably this work should happen in conjunction with a review/refactor of the MC functionality, so I'm not requesting a contributed PR yet.
* The other `Advanced_MC` examples involve Wang Landau algorithm which is more niche, and less accessible to typical Sunny users. Probably these examples should stay hidden until we have a strong use case that would motivate writing user-friend tutorials.
* The files in `KPM` were really only intended for us, as Sunny developers. Maybe they should just be removed so as not to confuse Sunny users who are browsing around?
* The `Plotting` directory was intended to hold helper functions that can help simplify the main tutorials. See, e.g., the how plots are made at the bottom of the [CP2 skyrmion tutorial](https://sunnysuite.github.io/Sunny.jl/stable/examples/06_CP2_Skyrmions.html)
In particular, I don't think it makes sense to expose these files as notebooks.
* [heisenberg_animation.jl](https://github.com/SunnySuite/Sunny.jl/blob/main/examples/extra/heisenberg_animation.jl) is sort of an odd-duck -- It's a nice illustration of creating interactive 3D graphics, but the animations don't easily translate to a web-browsable tutorial in the many Sunny docs.

kbarros:
Since `examples/extra` are not intended for notebook generation, I'm closing this PR. But in the future, we should revisit the question of making some of these examples more appropriate as full tutorials.

PR #349, opened 2025-01-06
Change convention for LSWT dynamical matrix

kbarros:
With this PR, the dynamical matrix $D$ is rescaled by a factor of 2 so that the quadratic bosonic Hamiltonian can be written,

$$ \mathcal{H} = \frac{1}{2} \Psi^\dagger D \Psi,$$

with $\Psi$ the HP bosons in Nambu spinor form. In this updated convention, the positive eigenvalues of $\tilde{I} D$ are directly the magnon excitation energies.

Apart from aesthetics, there is one real change in this PR: The effective LSWT `regularization` strength (with default of `1e-8`) is now weakened by a factor of 1/2. That is, the dynamical matrix is now shifted by `regularization` rather than `2 * regularization`. Consequently, the reference data for some "golden" tests must be updated. This modified regularization appears in the final commit of this branch.


ddahlbom (approved):
To my eye, this all looks good.

PR #360, opened 2025-03-25
Adapt test tolerances to true Float64 precision

ddahlbom:
Loosens the bounds on the "Kitchen Sink" LSWT dispersion test. Removes `g_tol` from call to `minimize_energy!` for dimers and also loosens the test bounds somewhat. 

Curiously, removing the `g_tol` and using defaults did *not* result in a sufficient optimum, even though explicitly setting `x_tol=1e-14` did produce sufficiently accurate results. This seems to indicate that the default settings (with `x_tol=0` and the secret `NaN` configuration) worked less well than having a finite `x_tol`. This is a bit odd.


kbarros:
These changes look good.

Note that `minimize_energy!` uses a default value of `x_tol=1e-12`, as shown here: https://github.com/SunnySuite/Sunny.jl/blob/01cacf57fb76ae49582987be5abb8279afe1530b/src/Optimization.jl#L133

Therefore it makes sense that you are getting about 12 digits of accuracy in the minimized spin magnitude.

I can't remember if there was a good reason for Sunny to select `x_tol=1e-12` specifically. It's nice to know that you were able to manually override `x_tol=1e-14` and achieve more accuracy.

test/test_lswt.jl
```diff
@@ -43,7 +43,7 @@
     @assert energy_per_site(sys) ≈ -328.38255
 
     # Verify that this is a local minimum of energy
-    @test norm(Sunny.proj.(Sunny.energy_grad_coherents(sys), sys.coherents)) < 1e-8
+    @test norm(Sunny.proj.(Sunny.energy_grad_coherents(sys), sys.coherents)) < 2e-8
```
kbarros, line 46: On my Mac, I verified we can get many more digits of accuracy here by selecting, say, `x_tol=1e-16` inside `minimize_energy!`.

test/test_lswt.jl
```diff
@@ -54,8 +54,8 @@
...
     data_golden = [0.00038665026, 0.0, 0.007231537047, 0.0, 0.008665001888, 0.0, 0.015340530772, 0.0, 0.0, 0.054200267869, 0.073309637341, 0.0, 0.005276014091, 0.0, 0.0, 0.026708463804, 0.0, 0.062332773978, 0.112027719874, 0.0, 0.031129659091, 0.0, 0.115569104228, 0.0, 0.0, 0.038994974901, 0.0, 0.031152494381, 0.029113142876, 0.0, 0.0, 0.00467095768, 0.000762927464, 0.000940556177, 0.0, 0.0, 0.00877327198, 0.0, 0.033263081608, 0.0, 0.066082588247, 0.0, 0.0, 0.059956690592, 0.334527718881, 0.007608761425, 0.0, 0.0, 0.0, 0.068446181755, 0.029366565904, 0.0, 0.0, 0.605679689903, 0.378667970539, 0.0, 0.858558512424, 0.0, 0.0, 0.290056738762, 0.349117318827, 0.0, 0.0, 0.955658357376, 0.0, 1.606184739756, 0.0, 0.201711065754, 0.181786511323, 0.0, 0.0, 0.313711125445, 0.769401980923, 0.0, 0.010377681107, 0.0, 0.151200728665, 0.327686235743, 0.0, 0.0]
-    @test isapprox(res.disp, disps_golden; atol=1e-9)
-    @test isapprox(res.data, data_golden; atol=1e-9)
+    @test isapprox(res.disp, disps_golden; atol=2e-8)
```
kbarros, line 57: This is a curious degradation -- I wonder if it would be resolved by tightening `x_tol` in the `minimize_energy!` call?

ddahlbom:
Thanks for looking this over. I can try refreshing the golden references with newly optimized data and compare across platforms later this week.

test/test_lswt.jl
ddahlbom: After explicitly setting `x_tol=1e-14` and refreshing the reference data, the tests pass on my Intel machines (Linux and Windows) as well as on my M2 Mac.

ddahlbom: The tests do not if the tolerance is relaxed back to `1e-12` (on my Linux workstation).

ddahlbom: They do however pass (with `x_tol=1e-12`) on the Mac.

quantumsteve:
> We might want to hold off a bit before modifying the thresholds. I'm discussing with `pkofod` about the Optim.jl tolerance parameters. Maybe if we could replace `x_tol` with, say, `x_reltol` then this could achieve higher accuracy on all platforms.

I'm seeing a warning about `x_tol` being deprecated. Do you already have a plan to update this?

```
 Warning: x_tol is deprecated. Use x_abstol or x_reltol instead. The provided value (1.0e-12) will be used as x_abstol.           
 @ Optim /opt/julia/packages/Optim/8dE7C/src/types.jl:110     
```

ddahlbom:
Thanks for pointing that out @quantumsteve. I haven't seen those warnings myself. I'll make sure my Optim is updated. It makes sense to change these to `x_abstol` if that's the case.

@kbarros Is there any new information from `pkofod`? I just started a new branch for Paddison's feature request, working off of the current main, and was reminded that I shouldn't expect the tests to pass on my Intel machine.

kbarros:
I don't think `x_reltol` or similar will be available in the near future, and this PR is probably the best strategy for now. I'll try to get to it within a couple days.

kbarros:
This PR is also needed to make tests pass on Julia 1.12 on Mac.

PR #387, opened 2025-05-29
Replace shadowed variable with assertion check

quantumsteve:


kbarros:
Hi @quantumsteve, thanks for your PR. Since it is a common pattern, I took this as an opportunity to unify the notation for iterating over "flattened" interaction arrays throughout the codebase. Please let me know what you think of the updated changes on this PR. CC @ddahlbom.

PR #393, opened 2025-06-13
Updates for Makie v0.23

kbarros:
The new `Makie.arrows3d` interface requires significant changes in Sunny, but overall leads to cleaner code. After these changes, it is not possible to maintain backwards compatibility with prior Makie versions.

Once https://github.com/MakieOrg/Makie.jl/issues/5059 lands it will be possible to reintroduce lighting settings to arrows (`diffuse` and `specular`) for increased color saturation.


ext/PlottingExt/PlotSpins.jl
```diff
@@ -81,11 +81,11 @@ function Sunny.plot_spins!(ax, sys::System; notifier=Makie.Observable(nothing),
...
-    arrowsize = 0.4a0
-    linewidth = 0.12a0
-    lengthscale = 0.6a0
-    markersize = 0.8linewidth
-    arrow_fractional_shift = 0.6
+    shaftradius = 0.06a0
+    tipradius = 0.2a0
+    tiplength = 0.4a0
+    lengthscale = 1.0a0
+    markersize = 1.6shaftradius
```
kbarros, line 88: These mappings are exact:
```jl
tipradius = arrowsize/2
shaftradius = linewidth/2
tiplength = arrowsize
```

The new total `lengthscale` matches the old `lengthscale` (previously only for the shaft) plus the `tiplength`.

ext/PlottingExt/PlotSpins.jl
```diff
@@ -128,39 +128,38 @@ function Sunny.plot_spins!(ax, sys::System; notifier=Makie.Observable(nothing),
...
+            notify.((vecs, pts, tipcolor))
             # isnothing(color) || notify(arrowcolor)
         end
 
         # Draw arrows
-        linecolor = (stemcolor, alpha)
-        Makie.arrows!(ax, pts_shifted, vecs; arrowsize, linewidth, linecolor, arrowcolor, diffuse=1.15, transparency=isghost)
+        shaftcolor = (stemcolor, alpha)
+        if !isempty(pts[])
+            Makie.arrows3d!(ax, pts, vecs; align=0.37, markerscale=1, tipradius, shaftradius, tiplength, tipcolor, shaftcolor, transparency=isghost) # FIXME: diffuse=1.15
```
kbarros, line 158: If `align = 0.5`, then the arrows would be exactly centered on `pts`. A value `align < 0.5` will shift the arrows in the direction of `dirs`. The choice `align = 0.37` was empirically determined by visual comparison with previous Sunny calculations.

kbarros:
With the updates, dipoles now vanish as below. Observe that cylindrical arrow shafts are now centered on the atom site (disregarding conical arrow tip).

image

kbarros:
After merging, we discovered some room for improvement in the [tutorial SW19](https://sunnysuite.github.io/Sunny.jl/stable/examples/spinw/SW19_Different_Ions.html).

The follow-up commit https://github.com/SunnySuite/Sunny.jl/commit/ee5f2db3697412118471881eb649910a6117873c improves the situation by tweaking the heuristic for the rescaling factor `c`. Specifically, the rescaling activates when the shaft length becomes less than half the tip length.

PR #401, opened 2025-06-24
Avoid storing duplicate logo files

kbarros:
Sunny logos are stored in the root `assets` directory. These will be copied into `docs/build/assets` at the beginning of each docs build.


kbarros:
An alternative implementation would link to `docs/src/assets/logo.svg` from within the project README. See the [BestieTemplate](https://github.com/JuliaBesties/BestieTemplate.jl/blob/85bf7805d60a717cd06dfb8614469b978667ecd6/README.md?plain=1#L2).

PR #411, opened 2025-07-23
Transparency in `plot_intensities` for bands data

kbarros:
This transparency achieves consistency with other Makie lines plots. Additionally, it significantly reduces the file size for PDF exports using the CairoMakie backend.

I also experimented with `Makie.image` instead of `Makie.heatmap`. This change is another way to significantly reduce the output PDF file size, because it stores the image in a compressed format. Conversely, `Makie.heatmap` generates PDFs that include one rectangle glyph per grid point, making compression much more difficult. Unfortunately, `Makie.image` has two drawbacks:

1. It is not compatible with `Makie.Resampler`, which we need for to avoid crashes for very large grid dimensions (https://github.com/MakieOrg/Makie.jl/issues/4950).
2. It introduces interpolation between grid cells, which obscures the discretenes of the calculated grid.

As an attempt to work around (2), `Makie.image` has an option `interpolate=false`, but unfortunately, this setting defeats the purpose. If interpolation is disabled, then the exported PDFs for `Makie.image` and `Makie.heatmap` have identical sizes. After some digging, I realized that this is an inherent limitation of PDFs -- there isn't a universally adopted standard for disabling interpolation on embedded images.


kbarros:
This PR also introduces an `interpolate` option to `plot_intensities`. Enable this to significantly reduce the size of exported PDFs that include heatmap data.

PR #430, opened 2025-10-09
struct EnergyMinimizationResult to store result from minimize_energy!

### LINKED ISSUE #407: `minimize_energy!` should return custom object carrying status information
The return value of `minimize_energy!` is currently an integer. If optimization was successful, it denotes the number of iterations. Otherwise, the return value is -1.

In the future, it would be better to return a custom struct with a `Base.show` method that provides more detail about the optimization status.

Sagar-Ramchandani:
Fixes #407 

This modifies `minimize_energy!` to return a struct `EnergyMinimizationResult` instead. 
The fields it contains are based on what was being computed in case of a non-convergence previously. 

I have also included an interface function `has_converged` to replace the previous check in the tests i.e. patterns such as `minimize_energy!(sys) > 0` are replaced by `has_converged(minimize_energy!(sys))`. 

Finally, this modification is *breaking* because the return type of `minimize_energy!` has changed, but this was expected to address this issue. 

Since this is my first PR to this project, please let me know if I have made a mistake somewhere and what I can do to correct it.

### DIFF
--- a/src/Optimization.jl
+++ b/src/Optimization.jl
@@ -6,23 +6,14 @@ end
 
 function Base.show(io::IO, ::MIME"text/plain", res::MinimizationResult)
     (; converged, iterations, data) = res
-    Δx = number_to_simple_string(Optim.x_abschange(data); digits=2)
-    g_res = number_to_simple_string(Optim.g_residual(data); digits=2)
+    Δx = number_to_simple_string(Optim.x_abschange(data); digits=3)
+    g_res = number_to_simple_string(Optim.g_residual(data); digits=3)
     if converged
         println(io, "Converged in $iterations iterations")
     else
-        printstyled(io, "Non-converged", underline=true)
-        println(io, " after $iterations iterations: |Δx|=$Δx, |∂E/∂x|=$g_res")
+        println(io, "Non-converged after $iterations iterations: |Δx|=$Δx, |∂E/∂x|=$g_res")
     end
 end
-
-"""
-    converged(res::MinimizationResult)
-
-Checks for convergence as marked by the return value of
-[`minimize_energy!`](@ref). For more detail, see the field `res.data`.
-"""
-Optim.converged(res::MinimizationResult) = res.converged
 
 
 # Returns the stereographic projection u(α) = (2v + (1-v²)n)/(1+v²), which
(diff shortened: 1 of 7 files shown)


src/Optimization.jl
```diff
@@ -144,15 +172,13 @@ function minimize_energy!(sys::System{N}; maxiters=1000, subiters=10, method=Opt
...
+            return MinimizationResult(true, cnt, res)
         end
 
         # Reset stereographic projection based on current state
         ns .= normalize.(iszero(N) ? sys.dipoles : sys.coherents)
         αs .*= 0
     end
 
-    f_abschange, x_abschange, g_residual = number_to_simple_string.((Optim.f_abschange(res), Optim.x_abschange(res), Optim.g_residual(res)); digits=2)
-    @warn "Non-converged after $maxiters iterations: |ΔE|=$f_abschange, |Δx|=$x_abschange, |∂E/∂x|=$g_residual"
```
kbarros: Is it OK that I removed the warning here?

src/Optimization.jl
```diff
@@ -1,3 +1,29 @@
...
+
+function Base.show(io::IO, ::MIME"text/plain", res::MinimizationResult)
+    (; converged, iterations, data) = res
+    Δx = number_to_simple_string(Optim.x_abschange(data); digits=2)
+    g_res = number_to_simple_string(Optim.g_residual(data); digits=2)
+    if converged
+        println(io, "Converged in $iterations iterations")
+    else
+        printstyled(io, "Non-converged", underline=true)
+        println(io, " after $iterations iterations: |Δx|=$Δx, |∂E/∂x|=$g_res")
```
kbarros: In the previous warning, we had also printed `|E|`, i.e. `Optim.f_abschange(data)`. That's the least useful of the three outputs, since `E` may be converged to tolerance $10^{-16}$ while `x` is only converged to tolerance $10^{-8}$. Either `|Δx|` or `|∂E/∂x|` is probably sufficient.

ddahlbom (approved):
This will be a nice improvement. Thanks for submitting this, Sagar.

I've often found myself wanting direct access to the `OptimizationResults`, particularly when trying to automate any kind of phase diagram study. This looks like a nice minimal way to make that easy.

PR #432, opened 2025-10-10
ql_slow : perform reversals in-place

Sagar-Ramchandani:
Modifies `ql_slow` to use in-place reversals. This requires Julia 1.6+ due to multi-dimensional `reverse!`. 

Note: The `ql_slow` function seems to not be in use (at least on main)? If not necessary, it might be better to delete it entirely.

### DIFF
--- a/src/MathBasics.jl
+++ b/src/MathBasics.jl
@@ -69,10 +69,10 @@ end
 # Returns the QL decomposition `Q, L = ql(A)` satisfying `Q * L ≈ A` with Q
 # orthogonal and L lower-triangular. 
 #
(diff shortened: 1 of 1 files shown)


kbarros:
Nice. I think we can omit the test because it's enough to document the behavior on this PR conversation. With this revised implementation it's no longer "slow" and we can call the function `ql`. The following works for me:

```jl
using Sunny
using LinearAlgebra
A = randn(5, 5)
(; Q, L) = Sunny.ql(A)
@assert Q * Q' ≈ I    # Q is orthogonal
@assert istril(L)     # L is lower-triangular
@assert Q * L ≈ A     # Correct decomposition
```

PR #481, opened 2026-03-16
Use cis(z) or cispi(z) when possible

quantumsteve:
```
Replace sin(π*<expr>) with sinpi(<expr>), cos(π*<expr>) with cospi(<expr>), and sincos(π*<expr>) with 
sincospi(<expr>). This is advantageous with regard to both accuracy and performance. As a particular example, to 
evaluate the sine function in degrees instead of radians, use sinpi(x/180.0). Similarly, the single-precision functions 
sinpif(), cospif(), and sincospif() should replace calls to sinf(), cosf(), and sincosf() when the function argument is of 
the form π*<expr>. (The performance advantage sinpi() has over sin() is due to simplified argument reduction; the 
accuracy advantage is because sinpi() multiplies by π only implicitly, effectively using an infinitely precise 
mathematical π rather than a single- or double-precision approximation thereof.)
```
https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/#math-libraries

### DIFF
--- a/src/EntangledUnits/EntangledSpinWaveTheory.jl
+++ b/src/EntangledUnits/EntangledSpinWaveTheory.jl
@@ -176,7 +176,7 @@ function intensities_bands(swt::EntangledSpinWaveTheory, qpts; kT=0)
         view(disp, :, iq) .= view(excitations!(T, H, swt, q), 1:L)
 
         for i in 1:nunits
-            Avec_pref[i] = exp(- im * dot(q_global, rs_global[i]))
+            Avec_pref[i] = cis(-dot(q_global, rs_global[i]))
         end
 
         Avec = zeros(ComplexF64, Nobs)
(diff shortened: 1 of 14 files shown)


kbarros:
Discussed with @quantumsteve on Slack. This PR was originally motivated by work on the [MAIQMag GPU fork](https://github.com/MAIQMag/Sunny.jl), but has negligible impact to CPU performance. Given that the GPU pathway is currently separate, there is not a strong justification for the degradation to code readability in the general case.

However:

- Changing `exp(i * x)` to `cis(x)` for real `x` might be important in performance-critical sections, even on the CPU. Let's consider that type of change on a per-case basis (especially if there are motivating benchmarks).
- For CUDA code specifically, @quantumsteve  noticed that changing `cis` to `cispi` in `precompute_dipole_ewald_at_wavevector_kernel` saves a register allocation. Once we get more code-sharing between CPU and GPU pathways, let's reconsider optimizations as guided by the profiler.

PR #488, opened 2026-04-23
Spin wave refactor

kbarros:
Calculate coefficients for the linear expansion of each observable $\hat{A}_q$ in the Nambu basis of Holstein-Primakoff bosons. This allows to unify observable-setup code-paths between usual linear-spin wave theory and the KPM/Lanczos solver.

The new implementation should also be a bit faster -- maybe 5-10% in some cases.

Observable-setup for the entangled units spin wave calculator does not currently fit within this revised framework. The intention is that `EntangledSpinWaveTheory.jl` will eventually be removed in favor of the ordinary spin wave theory pathway. This should be possible if each "displaced-subsite" gets its own observables within `MeasureSpec`, with appropriate `offsets`.


kbarros:
Superseded by #496.

Issue #343, opened 2024-12-11
Investigate `view_crystal` processing time


kbarros:
```jl
cryst = Sunny.pyrochlore_crystal()
@time view_crystal(cryst)
```

On my machine this takes 0.7 s. Then `@profview` shows about 25% of the time is in `basis_for_symmetry_allowed_couplings`. The cost of this function scales as follows:

```jl
@time reference_bonds(cryst, 1.0) # 0.063 s
@time reference_bonds(cryst, 2.0) # 0.42 s (note: 0.063 * 8 == 0.504)
@time reference_bonds(cryst, 3.0) # 3.32 s (note: 0.42 * 8 == 3.3s)
```

A cubic scaling wouldn't be surprising here. But maybe there are strategies to make this faster.

PR #493, opened 2026-06-08
Add shell-based truncated dipole cutoff

phyjswang:
Add `modify_exchange_with_truncated_dipole_dipole_by_shell!` for selecting a real-space dipole cutoff by neighbor distance shell. The cutoff is chosen midway between the requested shell and the next shell, then the existing distance-based modifier handles exchange replacement and reshaped-system transfer.

Reuse the shell grouping for the existing cutoff log message, export the new API, and add a focused chain-crystal test that verifies the first and second shell behavior.

### DIFF
--- a/src/Sunny.jl
+++ b/src/Sunny.jl
@@ -74,6 +74,7 @@ export Moment, System, Site, clone_system, eachsite, nsites, position_to_site, g
     set_exchange!, dmvec, enable_dipole_dipole!, set_field!, to_inhomogeneous, set_field_at!,
     set_vacancy_at!, set_onsite_coupling_at!, set_exchange_at!, set_pair_coupling_at!,
     symmetry_equivalent_bonds, remove_periodicity!, modify_exchange_with_truncated_dipole_dipole!,
+    modify_exchange_with_truncated_dipole_dipole_by_shell!,
     get_param, set_param!, get_params, set_params!, ParamSpec
 
 include("MagneticOrdering.jl")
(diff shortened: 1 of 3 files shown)


kbarros:
Thanks for your suggestion @phyjswang .

We discussed on Slack. Here is the short summary:

- When setting up interactions normally, Sunny does not usually print diagnostics, as this could become noisy for scripted workflows that repeatedly build systems.

- The relevant information is mostly contained in `print_symmetry_table`. It's a bit annoying that that function takes a distance, rather than a number of bonds. Partly that's historical (we don't want to break the public API), and partly that's because there are still some unresolved performance issues: #343. So this is the thing we actually need to fix.

- @phyjswang's specific need is to be able to interpret what a distance means in terms of atomic shells. They wanted this in the context measuring convergence of `modify_exchange_with_truncated_dipole_dipole!` towards `enable_dipole_dipole!`. For that purpose, here is a custom function that may do the job more simply:

```jl
function print_atom_shells(cryst, n)
    bonds = Sunny.reference_bonds_upto(cryst, n)

    total_cnt = 0
    for b in bonds
        println(b)
        println("Radius: ", Sunny.global_distance(cryst, b))
        coord = Sunny.coordination_number(cryst, b.i, b)
        total_cnt += coord
        println("Coordination number: ", coord)
        println("Cumulative atoms within radius: ", total_cnt)
        println()
    end
end
```

Example:

```
cryst = Sunny.kagome_crystal()
print_atom_shells(cryst, 5)

Bond(1, 2, [0, 0, 0])
Radius: 0.5
Coordination number: 4
Cumulative atoms within radius: 4

Bond(1, 2, [0, 1, 0])
Radius: 0.8660254037844386
Coordination number: 4
Cumulative atoms within radius: 8

Bond(3, 3, [1, 0, 0])
Radius: 1.0
Coordination number: 2
Cumulative atoms within radius: 10

Bond(1, 1, [1, 0, 0])
Radius: 1.0
Coordination number: 4
Cumulative atoms within radius: 14

Bond(1, 3, [-1, 0, 0])
Radius: 1.3228756555322954
Coordination number: 8
Cumulative atoms within radius: 22
```

If you just want the distance of the ``n``th shell, use:

```jl
Sunny.global_distance(cryst, Sunny.reference_bonds_upto(cryst, n)[end])
```

PR #494, opened 2026-06-12
Support `make_loss_fn` for `EntangledSystem`

kbarros:
This PR now includes many improvements to entangled units, fixing a set of bugs identified by Allen Scheie and Wolfgang Simeth.

- `make_loss_fn` supports EntangledUnits with simple wrappers.
- Fixes to intensity scale and RLU system for spin wave calculations with entangled units.
- Improved numerical robustness in tensor factorization, resolving occasional SVD errors.
- More informative error when a particular reshaping is not supported (would split an entangled unit across the periodic boundary).
- Additional tests to prevent regressions.

### DIFF
--- a/docs/src/versions.md
+++ b/docs/src/versions.md
@@ -3,6 +3,8 @@
 ## v0.9.2
 (In development)
 
+* Bug fixes and improvements to entangled units ([#494](@ref)).
+
 ## v0.9.1
 (Apr 25, 2026)
 
--- a/src/EntangledUnits/EntangledReshaping.jl
+++ b/src/EntangledUnits/EntangledReshaping.jl
@@ -22,31 +22,33 @@ function units_for_reshaped_system(reshaped_sys_uncontracted, esys)
     while length(new_atoms) > 0
         # Pick any site from list of new sites
         new_atom = new_atoms[1]
-        new_site = (1, 1, 1, new_atom) # Only need to define for a single unit cell, may as well be the first
-        new_position = position_at(reshaped_sys_uncontracted, new_site)
+        new_site = (1, 1, 1, new_atom) # Any representative cell works.
+        new_site_global = global_position_at(reshaped_sys_uncontracted, new_site)
 
-        # Find corresponding original atom number.
-        site = position_to_site(sys_uncontracted, position_at(reshaped_sys_uncontracted, new_site))
-        original_atom = site[4]
-        position_of_corresponding_atom = position_at(sys_uncontracted, (1, 1, 1, original_atom))
-        offset = new_position - position_of_corresponding_atom
+        # Map the reshaped site back to an atom in the original crystal.
+        new_site_in_original_basis = sys_uncontracted.crystal.latvecs \ new_site_global
+        original_site = position_to_site(sys_uncontracted, new_site_in_original_basis)
+        original_atom = original_site[4]
 
-        # Find the unit to which this original atom belongs.
+        # Find the original unit containing this atom.
         unit = units[findfirst(unit -> original_atom in unit, units)]
 
-        # Find positions of all atoms in the unit, map into the reshaped
-        # crystal, and reject any mapping that requires a nonzero lattice
-        # offset (indicating a split across crystallographic cells).
-        unit_positions = [position_at(sys_uncontracted, (1, 1, 1, atom)) for atom in unit]
-        new_unit_sites = map(unit_positions) do position
-            position_to_atom_and_offset(
-                reshaped_sys_uncontracted.crystal,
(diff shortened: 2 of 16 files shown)


ddahlbom (approved):
This all looks good to me. I tested it with the following simple example.

```
sing Sunny, LinearAlgebra, GLMakie

latvecs = lattice_vectors(1, 1, 2, 90, 90, 90)
positions = [[0, 0, 0], [0, 1/2 - 1e-4, 0]]
crystal = Crystal(latvecs, positions)
path = q_space_path(crystal, [[0, 1, 0], [1/2, 1, 0], [1, 1, 0]], 20)

J = 1
factor = 0.4rand()
J′ = factor*J
sys = System(crystal, [1 => Moment(s=1/2, g=2)], :SUN; dims=(1, 1, 1))
set_exchange!(sys, J, Bond(1, 2, [0, 0, 0]))
set_exchange!(sys, 1.0, Bond(1, 1, [1, 0, 0]), :J′ => 0)

labels = [:J′]
params = [J′]
set_params!(sys, labels, params)

esys = Sunny.EntangledSystem(sys, [(1, 2)])
randomize_spins!(esys)
minimize_energy!(esys)

eswt = SpinWaveTheory(esys; measure=ssf_trace(esys))
res = intensities_bands(eswt, path)
plot_intensities(res)


qs = path.qs
Es_ref = [[E] for E in res.disp[1,:]]

loss = make_loss_fn(esys, labels) do esys
    swt = SpinWaveTheory(esys; measure=ssf_trace(esys))
    println(get_params(esys, labels))
    res = intensities_bands(swt, qs)
    squared_error_bands(Es_ref, res)
end

guess = [0.25]
loss(guess);

import Optim

method = Optim.Fminbox(Optim.LBFGS())
options = Optim.Options(; g_tol=1e-8)
lower = [0.0]
upper = [0.5]
fit = Optim.optimize(loss, lower, upper, guess, method, options)
fitval = only(fit.minimizer)

println("
    J′ used = $J′
    J′ fit  = $fitval
    Δ = $(J′ - fitval) 
    err = $( abs(fitval - J′) / J′)
")
```
Only thing to worry about is ensuring that the search space is bounded to parameter values that yield a singlet. Otherwise you get Cholesky errors.

Should this be explicitly tested with an example that uses auto diff, or is auto diff only used for SCGA (and hence dipole systems)? As far as I could tell, the only requirement to make the CRC rules work would be an version of `clone_system` for `EntangledSystem`, and this does exist.

Thanks for putting this together.

EDIT: Should we ask Allen to try the branch out?

ddahlbom:
Following discussion with @kbarros, the code has been modified to rethrow the appropriate error such that `make_loss_fn` will return a large value when encountering a Cholesky factorization error. This was achieved by adding a `AbstractDirectSpinWaveTheory` type and eliminating some code duplication. In particular, the `excitations!` function used for ordinary `SpinWaveTheory` is reused by `EntangledSpinWaveTheory`.

A new problem has become apparent. For the example above, if bounds on the optimized parameters are removed, the following error is triggered triggered relatively often,
```
ERROR: AssertionError: norm(R * R' - I) < 1.0e-12
Stacktrace:
  [1] svd_tensor_expansion(D::Matrix{ComplexF64}, N1::Int64, N2::Int64)
```
i.e., the tensor decomposition of the interactions fails. 

After a couple hours of empirical investigation, a few things became clear:

1) Using the current assertion tests, the SVD fails for 3.5% of random J' values between 0 and 0.5. It's not clear there's a pattern in the J' values that fail. See figure below.
image

2) There are a small number of failures which are just above the assertion bound (~1e-11). The majority are massively over the bound, with `norm(R * R' - I)` being of the order of 0.1 or 1.

3) If I loosen the bound on the assertion to 1e-10, the failure rate drops about 2.5% and all failures massively exceed the bound (i.e. have values between 0.1 and 3). The distribution is shown below. It seems there are a few failure hotspots around J' = 0.15 and J'=0.34.

image


In brief, we are seeing that, for a small subset of parameters, the rotation matrix `R` constructed in the SVD decomposition fails to be orthogonal (by a significant margin). I can try to isolate a few specific examples of failure and try to see what `R` is doing.

ddahlbom:
I was getting failures in the test suite after our last iteration. The culprit was `diffnorm2(C, C') < 1e-12` assertion, which only works when `N1 == N2` (which ensures that `C` is square). The test suite contains mixed cases (e.g., biquadratic interaction between `s=1` and `s=2` sites). I switched this to a test for the realness of the matrix elements, i.e., the coefficients in c_{ij} H_i x H_j should always be real.

If this sounds reasonable, we could ask Allen or Wolfgang to try the branch, though judging by the recent email exchange it seems they've found their own workaround.

PR #496, opened 2026-07-03
Entangled units refactor

kbarros:
Support entangled units through the public `entangle_system` function. This is an extensive refactor of the previous prototype implementation (`EntangledSystem`). New capabilities:

* Better integration. An entangled system is an ordinary `System` in `:SUN` mode, with many behaviors automatically inherited.
* Unit specification that wraps the chemical cell is allowed. The `groupings` parameter accepts integer cell offsets for this purpose.
* Symmetry validation. It is a hard error if the "groupings" of atoms into units would break symmetry. To override, one can set `enforce_symmetry=false`. However, such an error may indicate accidental misspecification of the unit. For example, if the "cell offsets" were omitted, then symmetry analysis may suggest how to specify them correctly.
* Improved efficiency. The local dimension $N$ of an entangled unit can grow quickly in the number of unit parts, and the new implementation introduces algorithms that scale more favorably in $N$. Setup time is now $O(N^3)$, and the typical bottleneck sections of classical dynamics now scale like $O(N)$ when $N \lesssim 32$.
* Long-range dipole-dipole is supported through a generalization of Ewald summation that allows local entanglement.

A new bands-fitting tutorial illustrates an entangled dimer system: https://sunnysuite.github.io/Sunny.jl/previews/PR496/examples/12_Entangled_fitting.html


ddahlbom (approved):
We've been discussing as this it has progressed. Everything seems to be working as I would expect from my experiments. I've gone through this PR quickly, checking the parts of the code I am most familiar with. It looks good to me and feels appropriate to get it on main to have wider community exposure.

The additions for completeness, fixes for correctness, code integration and speedups for entangled units are great. It seems like this refactor opens the door to future improvements in a very nice way.

PR #505, opened 2026-08-27
Ewald Powder Speedup

danielguterding:
Reorganized the code for powder averages using dipole-dipole interactions evaluated via Ewald sums. The previous code re-calculated some heavy terms, which can actually be pre-computed and re-used.

In cases with 24 atoms per unit cell, the new code is about 60 times faster.

### DIFF
--- a/benchmarks/ewald_powder_cpu.jl
+++ b/benchmarks/ewald_powder_cpu.jl
@@ -1,0 +1,57 @@
+#!/usr/bin/env julia
+
+# Run from the repository root as:
+#   julia --project=. benchmarks/ewald_powder_cpu.jl [nq]
+
+using LinearAlgebra
+using Printf
+using Random
+using Sunny
+
+function fibonacci_sphere(n)
+    ϕ = (1 + √5) / 2
+    return [begin
+        z = 1 - 2(i - 0.5) / n
+        r = √max(0, 1 - z^2)
+        θ = 2π * (i / ϕ)
+        Sunny.Vec3(r*cos(θ), r*sin(θ), z)
+    end for i in 1:n]
+end
+
+function synthetic_dipolar_swt()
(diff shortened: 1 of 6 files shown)


kbarros:
Thanks for this contribution. Usually it helps to discuss big PRs first on the Sunny Slack workspace ([invite link](https://join.slack.com/t/sunny-users/shared_invite/zt-1otxwwko6-LzPtp7Fazkjx2XEqfgKqtA)). I've adapted some of the ideas here into #514. Would you like to be a coauthor on that one?

Reasons for not merging this PR:
* Code complexity increase (this is especially dangerous in the age of AI coding agents).
* Implementation should avoid fusing parts of the Ewald calculation into the spin-wave theory calculator. In particular, there are three "customers" of the Ewald tensor (spin wave theory in dipole mode, in SU(N) mode, and in spiral-calculator mode). The implementation of all three should remain parallel.
* Each $\mathbf{q}$ calculation is designed to be single-threaded. Any multi-threading should be lifted to the top level, e.g., via parallelization over independent $\mathbf{q}$ calculations. That's already been prototyped on the [`#threads`](https://github.com/SunnySuite/Sunny.jl/tree/threaded). That's not merged yet because, so far, we haven't seen big wins in our test applications. We can reconsider merging if you observe otherwise for certain important applications (tell us on Slack).

The follow-up #514 achieves the same level of single-threaded performance as this one, but uses a simpler implementation that optimizes _only_ the calculation of the Ewald  interaction tensor $A(\mathbf{q})$. There is even some additional speedup in #514 because it leverages Hermiticity of $A(\mathbf{q})$ to save work.

Side-comment: Sunny already has a built-in [powder_average](https://sunnysuite.github.io/Sunny.jl/stable/library.html#Sunny.powder_average) functionality.

PR #507, opened 2026-08-30
Sampled correlations convolution

### LINKED ISSUE #506: Add energy convolution to intensities calls from a SampledCorrelations
Currently, the functionality in `Broadening.jl` is only applicable to linear spin wave calculations. This should be extended to `SampledCorrelations` to enable the application of simple instrument resolution models.

ddahlbom:
Adds energy convolution functionality to LL calculations through a `kernel` keyword that behaves analogously to the `kernel` keyword used for LSWT intensity calls.

In particular, a `kernel` may be provided to `intensities(sc, ...)` calls but not `intensities_static(sc, ...)` calls. `intensities_static` does not accept a `kernel` keyword at all so will throw an error.

If a kernel is applied to an SC intensities call, it is mandatory that a sorted list of `energies` be provided as a keyword as well. The previous symbolic options (e.g., `:available`) do not work when a `kernel` keyword is given but are retained to maintain backward compatibility. That is, the symbolic options still work if no `kernel` is provided.

If a user selects a subset of the available energies via the `energies` keyword, the convolution is still performed on all the available energies (since convolution is non-local) but only the requested energies are returned.

As for the implementation, the convolution is done "by hand," so both stationary and non-stationary kernels are supported, as they are for LSWT. This avoids FFT boundary issues and is sufficiently fast for standard cases. 

The `kernel` option for `intensities(sc, ...)` is illustrated in the FeI2 LL tutorial now, but not in the CoRh₂O₄. It could be added if desired. The powder calculation for CoRh₂O₄ went from about 2.3 seconds to about 4.8 seconds on my workstation.


ddahlbom:
This addresses Issue #506

ddahlbom:
Investigating why the doc build suddenly stopped working...

kbarros:
Looks great David! The only I question I have is whether the kernel broadening is a bottleneck in practice. If so, I think we could precalculate the kernel matrix `kernel(ω0, ω)` independently of `q`. Then the convolution over all `q` becomes a matrix multiply.

We don't have that trick available for spin wave theory, because there the excitation energies `ω0` become `q` dependent. But in SWT,  the broadening slowdown is also less severe, because there are typically only a few bands to convolve.

Edit: It might be too much GC pressure to allocate `K = kernel(ω0, ω)` as a full matrix. That's not needed to get a big speedup. One might iterate over columns of `K`; these can be reused across all `q`.

kbarros (approved):

ddahlbom:
Thanks for the suggestion. I asked Claude to benchmark each approach (original, full matrix allocation, and vector allocation per omega0). In all but the single-q case, the alternatives you suggest are faster, and the vector buffer (kernel precalculated per \omega0) scales best. I committed the vector version.

| case | original | matrix-precompute | vector-per-ω0 |
|---|---|---|---|
| nω0=20000, nω=500, nq=1 | 0.045 s / 0.31 MiB | 0.061 s / 76.6 MiB (0.74x) | 0.038 s / 0.32 MiB (1.20x) |
| nω0=2000, nω=300, nq=100 | 0.271 s / 3.1 MiB | 0.045 s / 7.7 MiB (6.1x) | 0.017 s / 3.1 MiB (16.2x) |
| nω0=500, nω=200, nq=5000 | 2.261 s / 38.2 MiB | 0.260 s / 38.9 MiB (8.7x) | 0.120 s / 38.2 MiB (18.9x) |

PR #509, opened 2026-08-31
Sample correlations on selected q points

danielguterding:
Enable sampling of correlations on selected q values. This reduces memory usage for sampling the DSF on a q-path through the BZ.

### DIFF
--- a/src/SampledCorrelations/SampledCorrelationsQPoints.jl
+++ b/src/SampledCorrelations/SampledCorrelationsQPoints.jl
@@ -1,0 +1,335 @@
+mutable struct SampledCorrelationsQPoints{Q <: AbstractQPoints}
+    # 𝒮^{αβ}(q,ω) data and metadata for selected finite-grid q indices only.
+    const data              :: Array{ComplexF64, 5}              # (ncorrs × natoms × natoms × nq × nω)
+    const M                 :: Union{Nothing, Array{Float64, 5}}
+    const recipvecs         :: Mat3
+    const origin_crystal    :: Crystal
+    const Δω                :: Float64
+    const qpts              :: Q
+    const q_indices         :: Vector{CartesianIndex{3}}
+    const request_to_unique :: Vector{Int}
+    const request_qabs      :: Vector{Vec3}
+
+    # Observables
+    measure         :: MeasureSpec
+    const positions :: Vector{Vec3}
+
+    # Trajectory specs
+    const integrator :: AbstractIntegrator
+    const measperiod :: Int
+    nsamples         :: Int64
+
+    # Buffers and precomputed data.
+    const samplebuf  :: Array{ComplexF64, 6}
+    const corrbuf    :: Matrix{ComplexF64}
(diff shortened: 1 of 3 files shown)


kbarros:
Closing because the proper implementation of this idea will require an extended design discussion.

After chatting with @ddahlbom, we suspect the root problem is the data layout in `SampledCorrelations`. Specifically, its `data` field [scales quadratically in the unit cell size and linearly in number of correlations](https://github.com/SunnySuite/Sunny.jl/blob/main/src/SampledCorrelations/SampledCorrelations.jl#L3). An alternative design of `SampledCorrelations` might phase average and contract the data as it is being collected. I created #515 to document that idea. It would lead to significant space savings, and preserves the full grid of q-vectors, but brings complicated tradeoffs (we'd lose flexibility to query certain aspects of the data, and the relevant BZs must be specified up front).

@danielguterding, we don't have a concrete timeline for reducing Sunny's storage requirements. You indicated this is a serious limitation for your research. My recommendation for now is that you continue using your `SampledCorrelationsQPoints` helper functions on top of Sunny as it currently exists. Please comment here or on Slack what your memory requirements and calculation shapes are, and that will inform the design process of the eventual `SampledCorrelations` refactor.

kbarros:
@danielguterding Most users seem happy with the current storage layout, so we're trying to understand more about your use case. Reading between the lines, we're guessing that you have very large chemical cells, and the quadratic cost in `natoms` is the main problem. Is that right? Or something else?

danielguterding:
Yes, that is correct. Especially when sampling a 1D path through a big 3D supercell, my change makes the difference between feasible/unfeasible.

I am not saying this should be the standard way to calculate things for everybody.

kbarros:
Understood. But I'm trying to distinguish between "chemical cell size = crystal's `natoms`" and "finite lattice of chemical cells = system's `dims`". The current format is quadratic in `natoms`. Sunny is designed under the assumption that `natoms` is typically small. Do you have some reason for `natoms` to grow large?

One historical use case for large `natoms` was to model chemical inhomogeneities. But nowadays we have [`to_inhomogeneous`](https://sunnysuite.github.io/Sunny.jl/stable/library.html#Sunny.to_inhomogeneous), so that `natoms` can usually remain small.

danielguterding:
In my case the primitive cell has 24 magnetic atoms.

kbarros:
Thanks. How many Brillouin zones do you typically need?

danielguterding:
To get publication quality figures, I would use 40x40x40.

kbarros:
Thanks. I think you're saying that the system dims are $40^3$. But does your q-space path go much beyond [0, 1] in reciprocal lattice units (RLU)?

danielguterding:
Ok, sorry for misunderstanding you. No, I typically do not go beyond the first BZ.

kbarros:
Thanks, that's helpful. We'll think about the best way to support it.

danielguterding:
Thank you.

PR #510, opened 2026-09-06
Vectorize the Ewald dipole-dipole gradient contraction

riteshbhirud:
**Speedup: 2.76x on the isolated contraction loop, 1.58x on the full dipole-dipole gradient, measured on AMD EPYC 9654 (Zen 4, AVX-512), single-threaded.**

## Summary

The Fourier-space contraction in `accum_ewald_grad!` (`src/System/Ewald.jl`) is the dominant cost of the
long-range dipole-dipole gradient, which is evaluated on every gradient call during spin dynamics. The loop
indexes the long spatial axis with a middle `CartesianIndex` while keeping the two size-3 component indices
innermost, so it does not vectorize and spends a substantial fraction of its time on index arithmetic. The
FFTs themselves are a small part of the dipole-dipole cost; this contraction is the hot spot.

This reorders the loop so the three Fourier-space cell axes are collapsed into one linear axis carried by an
innermost `@simd` loop, with the component and sublattice indices moved outward. The `reshape`s share memory
with the originals, and the accumulation order into each element of `Fϕ` is unchanged.

## Benchmarks

Hardware: AMD EPYC 9654 (Zen 4, AVX-512), single-threaded, Julia 1.12.

Isolated contraction loop, at the Fourier-space array sizes produced by a 10x10x10-cell diamond lattice with
8 sublattices (minimum over 30 samples):

| | time |
|---|---|
| before | 1600.8 us |
| after | 580.7 us |
| speedup | **2.76x** |

End-to-end gradient (`set_energy_grad_dipoles!`) on the same system with dipole-dipole enabled, which is
dominated by this contraction:

| | time |
|---|---|
| before | 4839.2 us |
| after | 3060.6 us |
| speedup | **1.58x** |

### DIFF
--- a/src/System/Ewald.jl
+++ b/src/System/Ewald.jl
@@ -204,9 +204,14 @@ function accum_ewald_grad!(∇E, dipoles, @nospecialize(sys::System))
     # performance degrades by ~50%
     fill!(Fϕ, 0.0)
     (_, m1, m2, m3, na) = size(Fμ)
-    ms = CartesianIndices((m1, m2, m3))
-    @inbounds for j in 1:na, i in 1:na, m in ms, α in 1:3, β in 1:3
-        Fϕ[α,m,i] += conj(FA[α,β,m,i,j]) * Fμ[β,m,j]
+    nspatial = m1 * m2 * m3
+    Fϕr = reshape(Fϕ, 3, nspatial, na)
+    Fμr = reshape(Fμ, 3, nspatial, na)
+    FAr = reshape(FA, 3, 3, nspatial, na, na)
+    @inbounds for j in 1:na, i in 1:na, α in 1:3, β in 1:3
+        @simd for s in 1:nspatial
+            Fϕr[α, s, i] += conj(FAr[α, β, s, i, j]) * Fμr[β, s, j]
+        end
     end
 
     # Inverse Fourier transform to get ϕ in real space


src/System/Ewald.jl
```diff
@@ -204,9 +204,14 @@ function accum_ewald_grad!(∇E, dipoles, @nospecialize(sys::System))
...
     fill!(Fϕ, 0.0)
     (_, m1, m2, m3, na) = size(Fμ)
-    ms = CartesianIndices((m1, m2, m3))
-    @inbounds for j in 1:na, i in 1:na, m in ms, α in 1:3, β in 1:3
-        Fϕ[α,m,i] += conj(FA[α,β,m,i,j]) * Fμ[β,m,j]
+    nspatial = m1 * m2 * m3
+    Fϕr = reshape(Fϕ, 3, nspatial, na)
+    Fμr = reshape(Fμ, 3, nspatial, na)
+    FAr = reshape(FA, 3, 3, nspatial, na, na)
+    @inbounds for j in 1:na, i in 1:na, α in 1:3, β in 1:3
```
quantumsteve, line 211: would this be clearer if we separate α and β loops and place them inside the s loop?

quantumsteve: Rephrasing this, is the existing order of indices adding runtime?

riteshbhirud: Thanks for the review. I benchmarked this to answer your question directly.

The index ordering has a large, hardware-dependent effect. I tested the original, my reordered version, and your suggested ordering (α,β inside the s loop) as isolated loops.

On AMD EPYC 9654 (Zen 4), which is the hardware I had when I opened the PR, the reorder is about 2.76x faster than the original.

On an Intel Xeon Gold 5220 (Cascade Lake), the reorder is slower than the original. That machine was heavily loaded, so to rule out memory-bandwidth contention I ran a size sweep: even when the arrays fit entirely in L2 cache, where bandwidth contention is irrelevant, the reorder is about 1.7x slower than the original. So it is a genuine codegen and execution penalty on Intel, not a busy-machine artifact, and it grows at larger memory-bound sizes. The α,β-inside ordering is also slower than the original there.

So to answer directly: the existing order is not adding runtime on Intel, it is actually the fastest there. The speedup I measured was specific to the Zen 4 codegen, where the original happens to vectorize poorly. Since this regresses a very common CPU, I don't think the reorder is a good portable change.

kbarros:
Hi Ritesh, thanks for your contribution. Could you please post your benchmark script on this PR thread?

PR #514, opened 2026-09-13
Accelerate spin-wave calculations with Ewald

kbarros:
Significantly accelerate spin-wave theory when long-range dipole-dipole interactions are present. Follow-up to #505 authored by @danielguterding.

If dipole-dipole is enabled, building the dynamical matrix $D(\mathbf{q})$ requires running the full Ewald summation algorithm for each $\mathbf q$. The cost of Ewald scales quadratically in the size of the magnetic cell. Formally, this is "small" compared to the cubic cost to diagonalize $D(\mathbf{q})$. In practice, however, the large prefactors of Ewald summation cause it to dominate the overall cost of the spin-wave calculation.

This PR significantly accelerates spin-wave theory + Ewald summation using these observations:
1. The expensive "real-space terms" (involving `exp` and `erfc`) can be precalculated and cached in the `SpinWaveTheory` calculator. This caching allows reuse across $\mathbf{q}$.
2. The dominant cost of the Fourier-space part arises from the phase factors $\exp(i 2 \pi (\mathbf{m} + \mathbf{q}_0) \cdot \mathbf{r})$. We had been calculating this exponential for all points on the grid of integer cell offsets $\mathbf{m}$. Instead, given each $(m^x, m^y, m^z)$, it can be assembled as a product of 1D phases $\exp(i 2 \pi (m^\alpha + q_0^\alpha) r^\alpha)$.
3. The Ewald interaction tensor has Hermitian symmetry that can be leveraged for ~2x savings.

When the spin wave calculation involves many wavevectors $\mathbf{q}$, these optimizations yield a significant win. I've observed 20x to 30x speedups for full intensities calculations in dipole mode, various system sizes, single-threaded.


danielguterding:
Very nice. Thank you. I would be grateful to be co-author here.

